PR #6, opened 2021-11-18
Fix/encrypt json rpc

maycon-mello:
There is a PR in the sdk as well.
https://github.com/docknetwork/react-native-sdk/pull/6/files


src/polkadot-test.js
```diff
@@ -1,97 +1,130 @@
...
-  dock.setAccount(testAccount);
+function main() {
+    // const data = encrypt('test');
+    // const decrypted = decrypt(data);
+}
 
-  // const data = dock.api.tx.balances.transfer('37GfhtNUJk1aJhXuGxNJsAGenteDBX3DTVAvuBZm49Kqc9wA', '10000');
-  // const fee = await data.paymentInfo(testAccount);
+main();
+// import Keyring from '@polkadot/keyring';
```
mike-parkhill, line 31: is there a plan to uncomment this code?

src/core/crypto.js
```diff
@@ -0,0 +1,27 @@
+
+import crypto from 'crypto';
+
+const algorithm = 'aes-256-cbc';
+
+export const DEFAULT_KEY = '776fe87eec8c9ba8417beda00b23cf22f5e134d9644d0a195cd9e0b7373760c1';
```
mike-parkhill, line 6: is there a plan to allow this key to be injected rather than hard-coding it?

maycon-mello: I can  setup some env variables on the sdk, let me check

src/polkadot-test.js
maycon-mello: I think we can remove this file, I've been using for some testing.

src/core/crypto.js
```diff
@@ -3,9 +3,11 @@ import crypto from 'crypto';
 
 const algorithm = 'aes-256-cbc';
 
-export const DEFAULT_KEY = '776fe87eec8c9ba8417beda00b23cf22f5e134d9644d0a195cd9e0b7373760c1';
+export const DEFAULT_KEY = process.env.ENCRYPTION_KEY || process.env.REACT_APP_ENCRYPTION_KEY;
 export const SECURE_JSON_RPC = true;
 
+console.log(`encryption key`, DEFAULT_KEY);
```
mike-parkhill, line 9: probably shouldn't be logging this

maycon-mello: thanks

mike-parkhill (approved):

PR #10, opened 2022-01-07
new method to disconnect dock api

maycon-mello:


src/services/dock.js
```diff
@@ -136,6 +136,9 @@ export default {
       isDockReady = true;
       return result;
     },
+    async disconnect(...params) {
```
SamHellawell, line 139: should we set isDockReady to false?

maycon-mello: Yes

SamHellawell (approved):

PR #11, opened 2022-01-31
Feat/wallet sdk validations

maycon-mello:


packages/core/lib/core/validation.js
```diff
@@ -13,7 +13,7 @@ export function isNumberValid(v: any) {
 }
 
 export function assertKeyType(type) {
-  assert(KeypairTypes.find(t => t === type), `Invalid keypair type ${type}. Expected on of ${KeypairTypes.join(',')}`);  
+  assert(!!KeypairTypes.find(t => t === type), `Invalid keypair type ${type}. Expected on of ${KeypairTypes.join(',')}`);
```
mike-parkhill, line 16: linting and I'm not sure `Expected on of ${xx}` makes sense grammatically.

packages/core/lib/modules/accounts.js
```diff
@@ -188,23 +188,23 @@ export class Accounts {
     assert(!accountExists, Errors.accountAlreadyExists);
 
     if (json) {
-      const pair = await KeyringRpc.addFromJson(json, password);
+      const pair = await keyringService.addFromJson(json, password);
 
       assert(pair && pair.type, 'invalid keypair');
 
-      keyPairType = pair.type;
+      type = pair.type;
```
mike-parkhill, line 195: why the rename? Isn't keypairtype more descriptive?

mike-parkhill (commented):
what's happening with getting rid of the *rpc wrappers in the latest commits?

packages/core/lib/index.js
```diff
@@ -4,7 +4,6 @@ import rpcServer from './rpc-server';
 
 console.log('testing');
```
mike-parkhill, line 5: we should probably clear this log entry?

packages/core/lib/modules/wallet-backup.test.js
```diff
@@ -1,16 +1,7 @@
...
-  noDocumentsFound,
-  noAccountsFound,
-} from './wallet-backup';
-import {walletService} from '../services/wallet/service';
-import {getPromiseError} from '../services/test-utils';
-import emptyBackup from '../test/fixtures/backup-files/empty-json.json';
-import emptyBackup2 from '../test/fixtures/backup-files/wallet-backup-1.json';
+import { getPromiseError } from '../services/test-utils';
+import { walletService } from '../services/wallet/service';
 import backup11 from '../test/fixtures/backup-files/wallet-backup-11.json';
```
mike-parkhill, line 3: can we name the backup files to describe the scenario? e.g. `wallet-backup-empty-file.json`

packages/react-native/lib/message-handler.test.js
```diff
@@ -60,22 +62,22 @@ describe('Message handler', () => {
     expect(eventHandler._handleRpcResponse).toBeCalled();
   });
 
-  it('expect to handle json-rpc-request event', async () => {
-    const event = createTestEvent('json-rpc-request');
-    await eventHandler.handleEvent(event);
-    expect(eventHandler._handleRpcRequest).toBeCalled();
-    expect(rnRpcServer.receive).toBeCalled();
-  });
+  // it('expect to handle json-rpc-request event', async () => {
```
mike-parkhill, line 65: why are these commented out?

packages/core/lib/modules/accounts.js
maycon-mello: type is used by polkadot js library, trying to use the same parameter name

packages/core/lib/core/validation.js
maycon-mello: It should be `Expected one of ${xx}`. Let me fix that, is that better? Lemme know if you have any other sugestion

packages/core/lib/modules/accounts.js
mike-parkhill: Okay, that makes sense.

packages/core/lib/modules/wallet.test.js
```diff
@@ -45,6 +46,8 @@ describe('ApiModule', () => {
   it('Expect to import wallet from backup file', async () => {
     const password = 'test';
 
+    await wallet.close();
```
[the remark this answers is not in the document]
maycon-mello: This test case will create a new wallet from a backup file. But there is a wallet created already, being used by previous test cases, so it's required to close the wallet. There is a afterAll already, which will close the wallet as well.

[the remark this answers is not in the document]
maycon-mello: But just when all the tests are done

packages/core/lib/modules/wallet-backup.test.js
maycon-mello: fixed

packages/core/lib/modules/wallet.test.js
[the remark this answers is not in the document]
mike-parkhill: Do the other tests share data, or just the wallet? How much overhead does opening and closing the wallet for each test add?

This is probably fine as long as tests remain sequential and this test always runs after all the others, but it is baking some assumptions into the test flow that may or may not be clear to other devs coming in here.

scripts/test-teardown-globals.js
```diff
@@ -1,5 +1,5 @@
 module.exports =  () => {
-  if (process.env.CI) {
+  // if (process.env.CI) {
```
mike-parkhill, line 2: why are these commented out?

packages/core/lib/modules/wallet.test.js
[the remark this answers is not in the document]
mike-parkhill: Can you add a comment for this test that it needs to run last because it needs to setup its own wallet? Once you've done that I think I'm good.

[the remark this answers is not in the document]
maycon-mello: I've pushed a code refactoring. Now it will group related tests and make sure the wallet gets closed properly for each group.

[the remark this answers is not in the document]
mike-parkhill: I like that grouping better.

PR #15, opened 2022-03-29
Credentials storage

maycon-mello:


packages/core/package.json
```diff
@@ -17,6 +17,7 @@
         "bignumber.js": "^9.0.1",
         "react-native-keychain": "^8.0.0",
         "realm": "^10.10.1",
+        "axios": "^0.25.0",
```
mike-parkhill, line 20: we're using `isomorphic-unfetch` in Dock Certs. Is there an advantage to one over the other? Any value in being consistent?

packages/credentials/lib/credentials.test.js
```diff
@@ -0,0 +1,89 @@
...
+      add: jest.fn().mockReturnValue({
+        id: 1,
+        value: testCredential,
+      }),
+    };
+    const credentials = new Credentials({wallet});
+
+    it('Expect to download credential', async () => {
+      const url =
+        'https://creds.dock.io/1d28317eb63495340414fb11346d5b7f5fd50b65aa06c8064d88ec3ec993a29b?p=dGVzdA%3D%3D';
```
mike-parkhill, line 76: at some point we might need to switch to a locally spun up test web server so we're not overloading production, but for now this is probably fine.

mike-parkhill (commented):
do backups include credentials? Or will they?

maycon-mello:
> do backups include credentials? Or will they?

Yes, credentials are stored in the wallet as documents. So it will be included/restored on backups as we have for accounts.

packages/core/package.json
maycon-mello: Axios seems to have better tooling to mock requests, also reduces the amount of code required to handle json responses for example. I've never tried `isomorphic-unfetch` on react-native, but axios is supported without issues

mike-parkhill (approved):

PR #17, opened 2022-05-12
@docknetwork/wallet-sdk-dids

osinakayah:


packages/dids/src/did.methods/did.key.manager.ts
```diff
@@ -0,0 +1,48 @@
...
+  public static getInstance(): IDidManager {
+    if (!DidKeyManager.instance) {
+      DidKeyManager.instance = new DidKeyManager();
+    }
+
+    return DidKeyManager.instance;
+  }
+
+
+  async saveDID(did: string): Promise<boolean> {
```
SamHellawell, line 32: you need to save the whole DID JSON-LD document, not just a string. you also need to save the associated keypairs

packages/dids/src/did.methods/did.key.manager.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  private constructor() {
+    this.wallet = Wallet.getInstance();
+  }
+  async createDID(options = {}): Promise<string> {
+    const didDocument = await driver().generate();
+    return didDocument.id.toString()
+  }
+
+  async resolveDID(did: string): Promise<any> {
```
SamHellawell, line 19: this method should first check if the document exists in the wallet already or not as it can be used as a cache

packages/dids/src/did.methods/did.key.manager.ts
```diff
@@ -0,0 +1,48 @@
...
+export class DidKeyManager  implements IDidManager{
+  private static instance: IDidManager;
+
+
+  private wallet;
+
+  private constructor() {
+    this.wallet = Wallet.getInstance();
+  }
+  async createDID(options = {}): Promise<string> {
```
SamHellawell, line 14: creating a did string isnt really very useful, in this case itd be better to return the whole document

packages/dids/src/did.methods/did.key.manager.ts
SamHellawell: could make sense if we only used the did:key method, but to support more in the future makes sense to just store the documents so we dont have to ping the chain every time. it will also save the need to have to do expensive crypto operations on the app side through RPC just to view DID info

SamHellawell: please note though that you should not store the private keys in the document. this library generates keys and exposes the PKs inside the DID document when using generate. we will be sharing these DID documents between agents (wallets) at some point in the future, so we cant expose PK info. thats why they should be stored like account keys are currently in a separate document with an ID that can be referenced

packages/dids/src/did.methods/did.key.manager.ts
```diff
@@ -0,0 +1,52 @@
...
+    if (!DidKeyManager.instance) {
+      DidKeyManager.instance = new DidKeyManager();
+    }
+    return DidKeyManager.instance;
+  }
+
+
+  async saveDID(didDocument: any): Promise<DIDResponse> {
+    const doc = await this.wallet.add({
+      type: 'DID',
```
SamHellawell, line 29: this isnt the correct document format for a DID resolution. it should be like: https://w3c-ccg.github.io/universal-wallet-interop-spec/#DIDResolutionResponse

packages/dids/src/index.ts
```diff
@@ -0,0 +1,7 @@
+
+import { DidKeyManager } from './did.methods/did.key.manager';
+
+
+DidKeyManager.getInstance().saveDID('did:key:z6MkwdAManAKt5L5tp1BChkzfH1vESsdUbpnBcsnedsqqPRV').then((res)=>{
+  console.log(res)
```
maycon-mello, line 6: Is this required?

packages/dids/src/did.methods/did.key.manager.ts
osinakayah: Yeah. "content" property has the actual document

SamHellawell: which isnt correct

PR #18, opened 2022-05-18
Wallet did

osinakayah:


packages/dids/src/index.js
```diff
@@ -0,0 +1,61 @@
+import {Ed25519VerificationKey2020} from '@digitalbazaar/ed25519-verification-key-2020';
+import {CryptoLD} from 'crypto-ld';
+const didKeyDriver = require('@digitalbazaar/did-method-key').driver();
+
+export const DIDKeyManager = (function () {
+  let wallet;
+
+  const createDidKeyPair = async () => {
```
SamHellawell, line 8: we need to use the existing polkadot keypairs and their generation (exaclty the same as used for accounts)

SamHellawell: this class shouldnt even care about generating keypairs at all

packages/dids/src/index.js
```diff
@@ -0,0 +1,61 @@
...
+    const cryptoLd = new CryptoLD();
+    cryptoLd.use(Ed25519VerificationKey2020);
+    return cryptoLd.generate({type: 'Ed25519VerificationKey2020'});
+  };
+  const saveDiDKeyPair = keyPair => {
+    return saveToWallet({
+      type: 'KEY',
```
SamHellawell, line 15: this class shouldnt be saving keypairs, it should ideally take a key document from the wallet to produce the DID

packages/dids/src/index.js
osinakayah: > we need to use the existing polkadot keypairs and their generation (exaclty the same as used for accounts)
@SamHellawell 

Please do you have a reference to where we did this, probably in the dock certs or dock app.
Because from my understanding, we generate accounts from mnemonics in the dock app.

SamHellawell: the accounts are keypairs generated from a mnemonic seed. this is done in the dock app somewhere, but im not sure where. @maycon-mello can maybe help with this. essentially we just need to take a keypair document and get it to a method to format that into a did document (we need to do it this way to support did:dock later)

maycon-mello: we should be able to resolve the existing keypair document. Maybe could create something like `getDIDDocument(account: string)` passing the account id as arguments

PR #19, opened 2022-05-19
DID manager

osinakayah:


packages/dids/src/index.js
```diff
@@ -0,0 +1,34 @@
+import {v4 as uuidv4} from 'uuid';
+const didKeyDriver = require('@digitalbazaar/did-method-key').driver();
```
SamHellawell, line 2: should be able to use import here instead of require

packages/dids/tests/didmanager.test.js
```diff
@@ -0,0 +1,23 @@
...
+    const {didResolution} = await DIDKeyManager.createDID({});
+    expect(didResolution).toHaveProperty('correlation');
+    expect(didResolution).toHaveProperty('@context');
+    expect(didResolution).toHaveProperty('id');
+    expect(didResolution).toHaveProperty('type', ['DIDResolutionResponse']);
+    expect(didResolution).toHaveProperty('didDocument');
+    expect(didResolution.correlation.length).toBe(0);
+
+    console.log(didResolution);
+    expect(didKeyDriver._keyPairToDidDocument).toBeCalledWith({
```
SamHellawell, line 16: this isnt a valid test, no need to spy that this method gets called. should be able to check that it actually generates and returns a did:key document

packages/dids/tests/didmanager.test.js
```diff
@@ -0,0 +1,23 @@
+import {DIDKeyManager} from '../src';
+const didKeyDriver = require('@digitalbazaar/did-method-key').driver();
+
+describe('DID module', () => {
+  it('Create DID', async () => {
+    const keyDoc = {};
+    const {didResolution} = await DIDKeyManager.createDID({});
```
SamHellawell, line 7: passing empty object here doesnt allow the function to be tested properly

packages/dids/src/index.js
```diff
@@ -0,0 +1,34 @@
...
+    const {didDocument} = await didKeyDriver._keyPairToDidDocument({
+      keyPair: {
+        ...keyDoc,
+        keyPair: keyDoc,
+      },
+    });
+
+    const expiryDate = new Date();
+    expiryDate.setFullYear(expiryDate.getFullYear() + 1000);
+    const didResolution = {
```
SamHellawell, line 19: to me it makes more sense for this method to return the document, and you can export another method to get a did resolution document from any did document (future proofing)

packages/dids/src/index.js
```diff
@@ -0,0 +1,34 @@
+import {v4 as uuidv4} from 'uuid';
+const didKeyDriver = require('@digitalbazaar/did-method-key').driver();
+
+const DID_DEFAULT_CONTEXT = [
+  'https://w3id.org/wallet/v1',
+  'https://w3id.org/did-resolution/v1',
+];
+export const DIDKeyManager = (function () {
```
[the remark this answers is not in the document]
maycon-mello: another option would be to use a class here, similar to the credentials management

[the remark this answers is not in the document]
maycon-mello: but I agree that just for one method would make more sense to just `export keypairToDidKeyDocument(...){ `

maycon-mello (changes requested):
added some comments

packages/dids/src/index.js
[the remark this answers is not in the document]
osinakayah: I wrapped it because we might actually have more functions and I did not want to expose all of them.
It is just a pattern to make some stuffs private/public without actually using modifiers.

packages/dids/tests/didmanager.test.js
osinakayah: I had to mock it, that is why i had to check if it was called. I can't actually create a did:key in test env

maycon-mello (approved):

packages/dids/src/index.js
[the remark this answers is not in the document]
SamHellawell: pattern doesnt make sense because if its private you just dont export the method? you are here exporting basically like a static class whereby user has to do: `DIDManager.theMethod` instead of just `theMethod` which is cleaner, imo

packages/dids/tests/didmanager.test.js
SamHellawell: you can, why do you think you cant?

PR #22, opened 2022-05-30
Did rpc

osinakayah:


maycon-mello (approved):

packages/core/lib/services/dids/service-rpc.js
```diff
@@ -0,0 +1,17 @@
+import {RpcService} from '../rpc-service-client';
+import {serviceName} from './config';
+
+export class DIDServiceRPC extends RpcService {
+  constructor() {
+    super(serviceName);
+  }
+  keypairToDidKeyDocument(params) {
```
maycon-mello, line 8: Can we define some types for those parameters?

maycon-mello: And maybe add validation

maycon-mello: Similar to other rpc services we have

packages/core/lib/services/dids/index.test.js
```diff
@@ -0,0 +1,79 @@
...
+      correlation: ['4058a72a-9523-11ea-bb37-0242ac130002'],
+      controller: 'did:key:z6MkjjCpsoQrwnEmqHzLdxWowXk5gjbwor4urC1RPDmGeV8r',
+      type: 'Ed25519VerificationKey2018',
+      privateKeyBase58:
+        '3CQCBKF3Mf1tU5q1FLpHpbxYrNYxLiZk4adDtfyPEfc39Wk6gsTb2qoc1ZtpqzJYdM1rG4gpaD3ZVKdkiDrkLF1p',
+      publicKeyBase58: '6GwnHZARcEkJio9dxPYy6SC5sAL6PxpZAB6VYwoFjGMU',
+    };
+    const res = await service.keypairToDidKeyDocument({keypairDoc});
+    expect(DIDKeyManager.keypairToDidKeyDocument).toBeCalledWith(keypairDoc);
+    expect(res).toBeDefined();
```
mike-parkhill, line 28: can we assert anything specific about `res`?

packages/core/lib/services/dids/index.test.js
```diff
@@ -0,0 +1,79 @@
...
+        '#z6Mks8mvCnVx4HQcoq7ZwvpTbMnoRGudHSiEpXhMf6VW8XMg',
+      ],
+      capabilityDelegation: [
+        '#z6Mks8mvCnVx4HQcoq7ZwvpTbMnoRGudHSiEpXhMf6VW8XMg',
+      ],
+      keyAgreement: ['#z6LScrLMVd9jvbphPeQkGffSeB99EWSYqAnMg8rGiHCgz5ha'],
+    };
+    const res = await service.getDidResolution({didDocument});
+    expect(DIDKeyManager.getDidResolution).toBeCalledWith(didDocument);
+    expect(res).toBeDefined();
```
mike-parkhill, line 77: same here

packages/core/lib/services/dids/index.test.js
SamHellawell: yes, we should check that the DID ID is correct relative to the public key (its fingerprint)

PR #23, opened 2022-05-30
Fix/integration issues

maycon-mello:


packages/core/lib/modules/data-migration.test.js
```diff
@@ -0,0 +1,27 @@
...
+      'dock-wallet',
+      JSON.stringify(walletLegacyData),
+    );
+    wallet = await Wallet.create();
+    await wallet.ensureNetwork();
+  });
+
+  it('expect to create wallet with legacy data', async () => {
+    const accounts = await wallet.accounts.getAccounts();
+    expect(accounts.length).toBe(2);
```
mike-parkhill, line 21: should probably be verifiying that all the docs are now v0.2 too

packages/core/lib/modules/wallet.js
```diff
@@ -61,6 +66,16 @@ class Wallet {
     this.eventManager.registerEvents(WalletEvents);
     this.accounts = Accounts.getInstance({wallet: this});
 
+    global.walletInstances++;
+
+    setTimeout(() => {
+      if (global.walletInstances > 0) {
+        console.warn(
+          "Multiple wallet instances were created. If that's not intentional please check your code, and use the Wallet.getInstance() instead of creating a new instance",
```
mike-parkhill, line 74: what are the dangers of multiple wallets? How do I access a specific wallet if multiple are created?

packages/core/lib/test/fixtures/lagacy-wallet-schema.json
```diff
@@ -0,0 +1,42 @@
+{
```
mike-parkhill, line 1: typo nitpick: `lagacy` -> `legacy`

packages/core/lib/wallet/rpc-storage-interface.js
```diff
@@ -17,16 +17,18 @@ class RpcStorageInterface extends StorageInterface {
...
-      this.documents = JSON.parse(data);
+      data = await storageService.getItem(directory);
+      this.documents = typeof data === 'string' ? JSON.parse(data) : data;
 
       if (!this.documents) {
         this.documents = {};
       }
     } catch (err) {
-      Logger.info('error to retrieve data from rpc storage', err.toString());
+      Logger.error('error to retrieve data from rpc storage');
```
mike-parkhill, line 29: can we combine these into a single object rather than having three separate log lines? It will make parsing the logs later harder if it's on multiple lines.

packages/react-native/lib/index.js
```diff
@@ -15,13 +15,63 @@ import {Platform, View} from 'react-native';
...
+
+function getAccount(address, documents): AccountDetails {
+  const addressDoc = findDocument(address, documents);
+
+  if (!addressDoc) {
+    return null;
+  }
+
+  const correlation = findRelatedDocs(addressDoc, documents);
+  const currencyDoc = correlation.find(filterDocsByType('Currency'));
```
mike-parkhill, line 54: are we guaranteed that there will only ever be one `Currency` or `Mnemonic` doc? i.e. could `find` return an array?

packages/core/lib/modules/data-migration.test.js
maycon-mello: I've added a new tests for it, thanks

packages/react-native/lib/index.js
maycon-mello: Yes, these document are generated in the account creation. Also, the `find` method will return only one item and should not be an array

packages/core/lib/wallet/rpc-storage-interface.js
maycon-mello: fixed

packages/core/lib/modules/data-migration.test.js
```diff
@@ -0,0 +1,27 @@
...
+    unmockDockService = await mockDockService();
+    global.localStorage.setItem(
+      'dock-wallet',
+      JSON.stringify(walletLegacyData),
+    );
+    wallet = await Wallet.create();
+    await wallet.ensureNetwork();
+  });
+
+  it('expect to create wallet with legacy data', async () => {
```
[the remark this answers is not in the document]
maycon-mello: fixed

packages/core/lib/modules/data-migration.js
```diff
@@ -0,0 +1,47 @@
+import {WalletDocument} from '../types';
+import {Logger} from '../core/logger';
+import {Wallet} from './wallet';
+
+export const getWalletVersion = (docs: WalletDocument): String => {
+  const doc = docs.find((item: WalletDocument) => item.type === 'Metadata');
+  return (doc && doc.walletVersion) || '0.1';
```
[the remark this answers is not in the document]
maycon-mello: We have a metadata document in the wallet, it will be created in the migration step. So we don't need to have anything specific for each document

packages/core/lib/modules/data-migration.test.js
```diff
@@ -1,27 +1,54 @@
...
+      );
+      wallet = await Wallet.create();
+
+      await wallet.load();
+      await wallet.ensureNetwork();
+    });
+
+    it('expect to migrate accounts', async () => {
+      const accounts = await wallet.accounts.getAccounts();
+      expect(accounts.length).toBe(2);
```
mike-parkhill, line 28: how do we know the accounts were migrated?

packages/core/lib/services/dids/config.js
```diff
@@ -0,0 +1,25 @@
...
+  getDidResolution: ({didDocument}) => {
+    assert(typeof didDocument.id === 'string', 'invalid didDocument');
+  },
+};
+
+export type KeypairToDidKeyDocumentParams = {
+  keypairDoc: any,
+};
+
+export type GetDidResolutionParams = {
```
mike-parkhill, line 23: nitpick I noticed in Ossy's PR too. We seem to be inconsistent with naming of functions, variables etc. Sometimes we use `DID`, sometimes `Did` and sometimes `DiD`. We should pick one.

packages/core/lib/services/dids/index.test.js
```diff
@@ -0,0 +1,84 @@
...
+    };
+    const res = await service.keypairToDidKeyDocument({keypairDoc});
+
+    expect(DIDKeyManager.keypairToDidKeyDocument).toBeCalledWith(keypairDoc);
+    expect(res).toBeDefined();
+    expect(res).toHaveProperty('didDocument');
+    expect(res).toHaveProperty('keyDoc');
+  });
+
+  it('getDidResolution', async () => {
```
mike-parkhill, line 35: what are we expecting this test to prove?

packages/core/lib/modules/data-migration.test.js
maycon-mello: The .getAccounts() will return 0 items in case the migration didn't run. But i`ve also added another check ` expect(wallet.migrated).toBeTruthy();` so that can be used for future migrations too

packages/core/lib/services/dids/config.js
maycon-mello: I would go with `DID`. I can update that

packages/core/lib/modules/data-migration.test.js
maycon-mello: No, the new getAccounts() will be using an `Address` document and resolve the related documents such as `Currency`, `Keypair` and `Mnemonic`.  The old schema was using an `Account` document which is not valid for universal wallet, this document type will not be used anymore

maycon-mello: The getAccounts() here is used to make sure the wallet has the accounts we expected, and `wallet.migrated` will assert the migration was performed

packages/core/lib/services/dids/index.test.js
maycon-mello: might need some help from @osinakayah here

maycon-mello: I've renamed the test case

packages/core/lib/services/dids/index.test.js
```diff
@@ -0,0 +1,84 @@
...
+import {didService as service} from './service';
+import {validation} from './config';
+import {DIDKeyManager} from '@docknetwork/wallet-sdk-dids';
+
+describe('DID Service', () => {
+  it('ServiceRpc', () => {
+    assertRpcService(DIDServiceRPC, service, validation);
+  });
+
+  it('keypairToDidKeyDocument', async () => {
```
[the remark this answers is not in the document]
maycon-mello: I've renamed the test case

packages/core/lib/modules/data-migration.test.js
maycon-mello: then we should cover all the scenarios

packages/core/lib/services/dids/config.js
maycon-mello: I've pushed a refactor, would be nice to @osinakayah review too. To make sure thats ok. It will require an update in the app as well. I will push another PR in the app repository

mike-parkhill (approved):
Approved, but expect extra tests to follow.

packages/core/lib/modules/wallet.js
maycon-mello: Yes that sounds better, i know there are several implications. Just added the warning to prevent bugs, but throwing seems much better

PR #24, opened 2022-05-31
vs singing

maycon-mello:


mike-parkhill (commented):
this hooks into the QR scanning code that @SamHellawell already wired up, right?

maycon-mello:
Yes, it will be used by the app to sign the data. I'm looking to push an update in the app now, using this new method to sign the credential

PR #25, opened 2022-06-01
Signing credentials

osinakayah:


packages/core/lib/services/credential/service.js
```diff
@@ -0,0 +1,36 @@
...
+  rpcMethods = [
+    CredentialService.prototype.generateCredential,
+    CredentialService.prototype.signCredential,
+  ];
+  generateCredential() {
+    const vc = new VerifiableCredential();
+    vc.addType('DockAuthCredential');
+    return vc;
+  }
+  async signCredential(params) {
```
mike-parkhill, line 19: do these `params` need any validation?

osinakayah: Yes, that is what validation.signCredential(params); is for.

PR #29, opened 2022-06-02
fix/wallet balance issues

maycon-mello:
- release v0.2.1
- fixing issues with wallet balance

### DIFF
--- a/packages/core/lib/modules/accounts.js
+++ b/packages/core/lib/modules/accounts.js
@@ -78,9 +78,16 @@ export class Accounts {
       true,
     );
 
-    currency.value = balance;
+    console.log('fetching balance', balance);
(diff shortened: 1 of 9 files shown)


packages/core/lib/modules/accounts.js
```diff
@@ -78,9 +78,11 @@ export class Accounts {
       true,
     );
 
-    currency.value = balance;
+    if (currency.value !== balance) {
```
mike-parkhill, line 81: what was happening here?

maycon-mello: It could potentially create some loop, since every document update will push an event and there are hooks listening for updates which could potentially try to fetch balances. Just trying to avoid this to happen. It's also good for performance.

PR #33, opened 2022-06-14
fixing issues with json rpc resolvers

maycon-mello:


packages/core/lib/rpc-util.js
```diff
@@ -14,36 +45,24 @@ export function createRpcService(service) {
...
-        } catch (err) {
-          Logger.debug(`Error for ${methodName}`, err.toString());
-          throw err;
-        }
-      },
-    };
+  return methods.map(methodFn => {
+    return createMethodResolver({
+      methodFn,
+      methodName: methodFn.name,
```
maycon-mello, line 63: This method resolution works in the webview services. However the react native service, which will handle the storage, doesn't return the proper name.

packages/react-native/lib/rn-rpc-server/async-storage-service.js
```diff
@@ -1,11 +1,11 @@
 import AsyncStorage from '@react-native-async-storage/async-storage';
 
 export class StorageService {
-  rpcMethods = [
-    StorageService.prototype.setItem,
-    StorageService.prototype.getItem,
-    StorageService.prototype.removeItem,
-  ];
+  rpcMethods = {
+    setItem: StorageService.prototype.setItem,
```
maycon-mello, line 5: I've updated the react-native services to provide the method names, so that js minifiers will not change our resolution logic

packages/core/lib/rpc-util.js
maycon-mello: I'm keeping two ways to define json rpc services. Ideally we should refactor all the services later, just to avoid confusion and potential bugs

mike-parkhill (approved):
Let's follow this fairly quickly with some tests if we can.

PR #34, opened 2022-06-14
fixing subscan issues

maycon-mello:


packages/transactions/lib/transactions.js
```diff
@@ -182,9 +191,17 @@ export class Transactions {
     let data;
     let page = 0;
     do {
-      data = await fetchTransactions({address, page});
-      data.transfers.forEach(handleTransaction);
-      page++;
+      try {
+        data = await fetchTransactions({address, page});
+        data.items.forEach(handleTransaction);
+        Wallet.getInstance().eventManager.emit(
```
maycon-mello, line 197: for each page loaded the UI will be updated, so that we don't need to wait for all the pages to load. Ideally we should have pagination in the UI instead of fetching all the pages.

maycon-mello: Created a jira for that https://dock-team.atlassian.net/browse/DCKA-793

mike-parkhill (approved):

PR #40, opened 2022-06-23
data migration issues

maycon-mello:


packages/core/lib/modules/data-migration.js
```diff
@@ -1,11 +1,12 @@
 import {Logger} from '../core/logger';
+import {getStorage} from '../core/storage';
 import {Wallet} from './wallet';
 
 type MigrateParams = {
   wallet: Wallet,
 };
 
-const currentWalletCersion = '0.2';
```
mike-parkhill, line 8: how did I miss this typo before?

packages/core/lib/modules/data-migration.js
```diff
@@ -27,6 +28,11 @@ export async function migrate({wallet}: MigrateParams) {
 
   Logger.debug(`Wallet version ${version}`);
 
+  if (version !== currentWalletVersion) {
+    const snapshot = await getStorage().getItem(wallet.walletId);
+    await getStorage().setItem(`${wallet.walletId}-snapshot`, snapshot);
```
mike-parkhill, line 33: how do we want to handle this blowing up? Do we catch it and continue on? Let the app crash? But if we let it crash do we get into a state where it will crash forever?

maycon-mello: If it crashes, the `migrated` event will not be fired and the app will stay in the splash screen forever

maycon-mello: we will have sentry logs for that, but not sure if that would make sense to display the wallet UI using data in a wrong format

maycon-mello: would create a lot of other issues probably

packages/core/lib/modules/wallet.js
```diff
@@ -27,6 +29,19 @@ export const WalletEvents = {
   networkConnected: 'network-connected',
 };
 
+/**
```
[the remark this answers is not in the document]
mike-parkhill: what if instead of passing a bool we passed the path to the schema file that we want to load? Would support different schema versions that way as we go forward.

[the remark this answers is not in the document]
mike-parkhill: Ok, we can leave it for now but would be good to support different versions easily.

mike-parkhill (approved):

PR #47, opened 2022-07-07
fixing issue with send max tokens

maycon-mello:


packages/core/lib/services/substrate/service.js
```diff
@@ -47,7 +47,10 @@ export class SubstrateService {
       amount,
     );
     const paymentInfo = await extrinsic.paymentInfo(account);
-    return paymentInfo.partialFee.toNumber() / DOCK_TOKEN_UNIT;
+    return (
+      paymentInfo.partialFee.muln(110).div(BN_HUNDRED).toNumber() /
```
mike-parkhill, line 51: what's the 110 and BN_HUNDRED?

Can we get a test added that protects against this issue in the future?

maycon-mello: it's adding 10% to the actual fee, this is the logic used by polkadot fe

maycon-mello: Yes let me add a test to it

mike-parkhill: Can we add a constant like, FEE_ESTIMATION_BUFFER to make it clear what it is? Also could you achieve the same thing by doing a simple `muln(1.10)` and save an operation? It would be clearer to me as a decimal what you're doing

PR #49, opened 2022-07-08
fix/build issues

maycon-mello:
- adding missing packages to the rn module
- fixing build issue


maycon-mello:
I've added the build script in the CI

SamHellawell (approved):

PR #51, opened 2022-07-15
DID key creation

maycon-mello:


packages/react-native/lib/didHooks.test.js
```diff
@@ -0,0 +1,145 @@
...
+            fetchBalance: jest.fn(() => Promise.resolve(0)),
+          },
+          clear: jest.fn(() => (documents = [])),
+        },
+        documents,
+      };
+    }),
+  };
+});
+describe('DID Hooks', () => {
```
mike-parkhill, line 107: can we have some failure case tests too?

osinakayah: Sure.

packages/react-native/lib/didHooks.test.js
```diff
@@ -123,6 +123,20 @@ describe('DID Hooks', () => {
...
+  test('Create new DID with invalid params', async () => {
+    const {result} = renderHook(() => useDIDManagement());
+
+    await expect(
+      result.current.createKeyDID({
+        derivePath: '',
+        type: 'sr25519',
+        name: 'DID Name',
+      }),
+    ).rejects.toThrowError('sr25519 keypair type  is not supported.');
```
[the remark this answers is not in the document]
osinakayah: Yes, I am trimming that now.

mike-parkhill (commented):
Looks like the tests are failing to run, but the action isn't failing:

`FAIL packages/transactions/lib/transactions.test.js
  ● Test suite failed to run
    Cannot find module '@docknetwork/wallet-sdk-dids' from 'setup-tests.js'
      28 | jest.mock('@react-native-async-storage/async-storage', () => 'AsyncStorage');
      29 |
    > 30 | jest.mock('@docknetwork/wallet-sdk-dids', () => {
         |      ^
      31 |   const originalModule = jest.requireActual('@docknetwork/wallet-sdk-dids');
      32 |   const moduleFunctions = {
      33 |     keypairToDIDKeyDocument: jest.fn().mockResolvedValue({
      at Resolver.resolveModule (node_modules/jest-runtime/node_modules/jest-resolve/build/index.js:306:11)
      at Object.<anonymous> (setup-tests.js:30:6)`

mike-parkhill:
I'm almost more concerned that the action didn't fail than I am that the tests failed, so if you could take a look at both I'd appreciate it.

scripts/test-teardown-globals.js
```diff
@@ -1,5 +1 @@
-module.exports =  () => {
-  if (process.env.CI) {
-    process.exit(0)
-  }
-};
\ No newline at end of file
+module.exports = () => {};
```
mike-parkhill, line 1: do we still need this file if it's not really doing anything?

packages/core/lib/services/wallet/service.js
```diff
@@ -275,6 +276,27 @@ export class WalletService {
 
     return this.wallet.getStorageDocument({id}).then(doc => doc.content);
   }
+
+  async getDocumentsFromEncryptedWallet(params) {
```
mike-parkhill, line 280: we're good with errors bubbling out of this function?

osinakayah: Yes we are. The plan is to allow the caller handle the exception how they chose to.

mike-parkhill (approved):

PR #52, opened 2022-07-21
fixing issue with did tests

maycon-mello:


maycon-mello:
This is ready for review

mike-parkhill (commented):
What's happening with the deleted tests? Are they going to be replaced or are they no longer relevant?

Also, does this fix the Action not failing when the tests do?

maycon-mello:
> 

We had some tests which were not valid anymore, so had to remove those.

This PR also fixes the github action, it should not pass with failed tests

PR #54, opened 2022-07-28
fixing network issue

maycon-mello:


packages/core/lib/modules/network-manager.js
```diff
@@ -27,13 +27,6 @@ export const SUBSTRATE_NETWORKS = {
 };
 
 function getNetworkInfo(networkId): NetworkInfo {
-  console.log('getNetworkInfo', networkId);
```
mike-parkhill, line 30: I'm embarrassed that I missed that.

mike-parkhill (approved):

packages/core/lib/modules/network-manager.js
mike-parkhill: Maybe in the future throw a TODO or something in when adding debug code

maycon-mello: Me too 😅

PR #55, opened 2022-07-28
Dcka 828 did management

osinakayah:


packages/react-native/lib/didHooks.js
```diff
@@ -142,13 +142,41 @@ export function useDIDManagement() {
...
+  const exportDID = useCallback(
+    async ({id, password}) => {
+      const existingDoc = await wallet.getDocumentById(id);
+      if (existingDoc) {
+        const correlationDocuments = await wallet.resolveCorrelations(id);
+        const didKeypairDoc = correlationDocuments.find(doc => {
+          return !!(
+            doc &&
+            typeof doc.type === 'string' &&
+            doc.type.startsWith('Ed25519VerificationKey')
```
mike-parkhill, line 154: will this prevent us from exporting non-ED25519 DIDs?

osinakayah: Yes, it would.

mike-parkhill: Is that what we want to do? I would think we should export any Did regardless of the keys used.

maycon-mello (approved):

PR #56, opened 2022-07-29
did dock

maycon-mello:


osinakayah:
@maycon-mello 
Do we need a function for calculating the transaction fee needed for create a did:dock?

packages/core/lib/services/dids/index.test.js
```diff
@@ -99,6 +99,6 @@ describe('DID Service', () => {
 
   it('expect to register did dock', async () => {
```
osinakayah, line 100: Can we include another test case for when something goes wrong.
Like, if no Keypair was found, or if there is insufficient for the transaction fee.

osinakayah (changes requested):

maycon-mello:
> @maycon-mello Do we need a function for calculating the transaction fee needed for create a did:dock?

I can add one in a separate PR

osinakayah:
Some tests are failing.

maycon-mello:
The failed tests seems related to the substrate mock, I'm working to fix that. Also, its showing as passing in the CI again, but shouldn't

osinakayah (approved):

PR #60, opened 2022-08-05
fixing issues with js bundle

maycon-mello:


fixdeps.js
```diff
@@ -0,0 +1,10 @@
+const glob = require('glob');
+const fs = require('fs');
+console.log(glob);
+
+glob('node_modules/@polkadot/**/packageInfo.cjs', {}, function (er, files) {
+  console.log(files);
```
osinakayah: We should remove the console logs

maycon-mello: Sure, I've pushed already this commit https://github.com/docknetwork/react-native-sdk/pull/60/commits/7e7e2a513f263b2dd7bccd46af0e47e69b113ac1

osinakayah (approved):

PR #62, opened 2022-08-11
Add keydoc to correlations

osinakayah:


packages/react-native/lib/didHooks.test.js
```diff
@@ -534,7 +535,11 @@ describe('DID Hooks', () => {
       name: 'DID Name',
       didType: 'diddock',
     });
-    expect(walletResult.current.wallet.add).toHaveBeenCalled();
+    expect(didServiceRPC.generateKeyDoc).toHaveBeenCalledWith({
+      derivePath: '',
+      type: 'ed25519',
+    });
+    expect(walletResult.current.wallet.add).toHaveBeenCalledTimes(2);
```
mike-parkhill, line 542: should we inspect the wallet to ensure the right data is there?

osinakayah: The app kind of handles that already via the UI. It makes sure all the inputs are valid before the "confirm button" is enabled, and we have taken a snapshot of this.

mike-parkhill: but that's depending on an external application to do your testing. What if someone makes changes here and doesn't run the app tests?

osinakayah: Here, the wallet instance is mocked, so this is to test to make sure the wallet add function is called exactly how it should be called. The wallet instance is gotten from a different package "@docknetwork/wallet-sdk-core", that is the reason i mocked it.


The unit test that makes sure when the "add" function is called, the document is actually saved is here: 
https://github.com/docknetwork/react-native-sdk/blob/02c53c007667883ed5aac4a80c606b300288d9c5/packages/core/lib/modules/wallet.test.js#L33

Trying to do it here again, I think would be double testing.

mike-parkhill (approved):

PR #68, opened 2022-09-14
Dcka 1001 format currency

osinakayah:


packages/core/lib/core/format-utils.js
```diff
@@ -22,13 +22,12 @@ export const DOCK_TOKEN_UNIT = 1000000;
  * @returns string
  */
 export function formatCurrency(value, currency = 'USD'): string {
-  assert(!!value, 'value is required');
   assert(
     typeof value === 'number' || typeof value === 'bigint',
     'value must be a number or bigint',
   );
 
-  var formatter = new Intl.NumberFormat('en-US', {
```
maycon-mello, line 31: Should we get the locale from some state maybe? Wondering how we can switch this to use the phone defaults later on. Perhaps we should create some Intl manager file, or something to store the current locale selection

osinakayah: How about we pass the locale via the function parameter params? We can default it to 'en-US'

packages/core/lib/core/format-utils.test.js
```diff
@@ -31,11 +31,14 @@ describe('Format util', () => {
...
+
+    const result1 = formatCurrency(0);
+    expect(result1).toBe('$0.00');
   });
 
   it('format Date', () => {
     const date = new Date(1996, 11, 17);
     const result = formatDate(date);
-    expect(result).toBe('December 17, 1996, 12:00 AM');
+    expect(result).toBe('December 17, 1996 at 12:00 AM');
```
mike-parkhill, line 42: I'm always suspicious when I see date format tests showing a new result with no related code change

mike-parkhill (approved):

PR #73, opened 2022-10-11
fix issues with send max tokens

maycon-mello:


packages/core/lib/services/substrate/service.js
```diff
@@ -46,7 +46,8 @@ export class SubstrateService {
   async getFeeAmount(params: TransactionParams) {
     validation.getFeeAmount(params);
 
-    const {toAddress, fromAddress, amount} = params;
+    const {toAddress, fromAddress} = params;
+    const amount = getPlainDockAmount(params.amount).toNumber();
```
mike-parkhill, line 50: any chance `getPlainDockAmount()` could return something that can't be converted to a number? And if so, what's the result? Does it throw an exception? Or return 0 or -1 or something?

mike-parkhill (commented):
are you able to unit test this with different scenarios to make sure they all still work?

maycon-mello:
> are you able to unit test this with different scenarios to make sure they all still work?

It throws on polkadot-js api, and we are still mocking the api connection. So its not possible to cover this scenario properly

maycon-mello:
Related to this jira https://dock-team.atlassian.net/browse/DCKA-963 @mike-parkhill

packages/core/lib/services/substrate/service.js
maycon-mello: the `validation.getFeeAmount(params);` will throw in case the amount is not a number

maycon-mello: There is a check for `isNumberValid` on it

mike-parkhill: are we handling that exception gracefully in the UI already?

maycon-mello: Thanks, i've added this to dock-app repo https://github.com/docknetwork/dock-app/pull/372

mike-parkhill (approved):

PR #78, opened 2022-11-05
Request logger

osinakayah:


maycon-mello:
Can we have some method to clear the logs?

maycon-mello (approved):

PR #81, opened 2022-11-09
Use diddock as controller

osinakayah:


packages/react-native/lib/didHooks.js
```diff
@@ -5,7 +5,11 @@ import {useWallet} from './index';
 export function useDIDUtils() {
   const createDIDKeypairDocument = useCallback(async keypairParams => {
     const {type, derivePath} = keypairParams;
-    return await didServiceRPC.generateKeyDoc({type, derivePath});
+    return await didServiceRPC.generateKeyDoc({
```
maycon-mello, line 8: I think we can remove the `await`  here

osinakayah: Resolved

maycon-mello (approved):

PR #87, opened 2022-12-02
Feat/relay service

maycon-mello:


packages/core/lib/modules/wallet.js
```diff
@@ -274,7 +272,7 @@ class Wallet {
       throw new Error(
         'Wallet module timed out. Make sure the wallet is loaded, or you are not using multiple instances',
       );
-    }, 6000);
+    }, 30000);
```
mike-parkhill, line 275: that's a pretty big change - what's the reasoning for going that big?

packages/core/lib/services/relay-service/configs.js
```diff
@@ -0,0 +1,31 @@
+import assert from 'assert';
+import {isNumberValid} from '../../core/validation';
+
+export const validation = {
+  sendMessage({keyPairDoc, recipientDid, message}: SendMessageParams) {
+    assert(!!keyPairDoc, 'keyPairDoc is required');
+    assert(!!recipientDid, 'recipientDid is required');
+    assert(typeof message === 'string', 'invalid message');
```
mike-parkhill, line 8: would there be value to accepting an object and serializing it? Or are you deliberately leaving serialization to the caller?

packages/relay-service/lib/index.js
```diff
@@ -0,0 +1,60 @@
+import assert from 'assert';
+import axios from 'axios';
+import {generatePayload, toBase64} from './payloads';
+
+let serviceURL = 'https://relay.dock.io';
```
mike-parkhill, line 5: can we read this from a config? Thinking at some point we may want to be able to change it (test env, white label clients running their own, etc.)

packages/relay-service/lib/payloads.js
```diff
@@ -0,0 +1,58 @@
+import VerifiableCredential from '@docknetwork/sdk/verifiable-credential';
+import {cryptoWaitReady} from '@polkadot/util-crypto';
+import {getKeypairFromDoc} from '@docknetwork/universal-wallet/methods/keypairs';
+import {getSuiteFromKeyDoc} from '@docknetwork/sdk/utils/vc/helpers';
+
+// 1 year
+const DEFAULT_EXPIRATION = 86400 * 1000 * 365;
+
+export async function generatePayload(keyPairDoc, subject) {
```
mike-parkhill, line 9: should probably rename this to indicate what kind of payload it generates

packages/relay-service/lib/payloads.js
```diff
@@ -0,0 +1,58 @@
...
+      RelayAuthCredential: 'dk:RelayAuthCredential',
+      limit: 'dk:limit',
+      to: 'dk:to',
+      msg: 'dk:msg',
+    },
+  ]);
+
+  cred.setIssuanceDate(new Date().toISOString());
+  cred.setExpirationDate(
+    new Date(Date.now() + DEFAULT_EXPIRATION).toISOString(),
```
mike-parkhill, line 26: would it make sense to have to renew this more frequently?

mike-parkhill: in a similar vein, have we given any thought to de-authorizing the connection in general?

maycon-mello: I can handle it during this week

packages/relay-service/lib/index.js
```diff
@@ -0,0 +1,62 @@
+import assert from 'assert';
+import axios from 'axios';
+import {generateSignedPayload, toBase64} from './payloads';
+
+let serviceURL = process.env.RELAY_SERVICE_URL || 'https://relay.dock.io';
+
+const sendMessage = async ({keyPairDoc, recipientDid, message}) => {
+  assert(!!keyPairDoc, 'senderDid is required');
```
mike-parkhill, line 8: Message doesn't match the parameter name

mike-parkhill (commented):
One little nitpick

maycon-mello:
can you please check again @mike-parkhill

mike-parkhill (approved):

PR #88, opened 2022-12-07
Integrate BBS to wallet sdk

osinakayah:


maycon-mello (changes requested):
There is a lint issue to be fixed

packages/core/lib/services/credential/service.js
```diff
@@ -21,6 +23,15 @@ const resolver = new MultiResolver(
   new UniversalResolver('https://uniresolver.io'),
 );
 
+let bbsPlusPresentation;
```
mike-parkhill, line 26: why are we keeping this as state? Do we need it? If so, what happens if I generate two presentations?

maycon-mello (approved):

PR #90, opened 2022-12-19
feat: pex filters

maycon-mello:


packages/core/lib/services/credential/service.js
```diff
@@ -84,6 +89,28 @@ class CredentialService {
...
+      presentationDefinition,
+      credentials,
+      holderDid,
+    );
+
+    return result;
+  }
+
+  evaluatePresentation(params) {
+    const {presentation, presentationDefinition} = params;
```
mike-parkhill, line 105: do we need to worry about validating these params before passing them to pex?

same with function above

mike-parkhill: and what about tests?

maycon-mello: pex will validate that for us

mike-parkhill (approved):

PR #92, opened 2023-01-11
Check if password will be required to import credential

osinakayah:


packages/credentials/lib/credentials.test.js
```diff
@@ -100,5 +100,18 @@ describe('Credentials module', () => {
       const error = await getPromiseError(() => credentials.addFromUrl(url));
       expect(error.message).toBe('Invalid credential');
     });
+
+    it('Expect check if password is required', async () => {
+      const url =
```
SamHellawell, line 105: are these requests mocked? if not they should be, cant guarantee this url will exist in future and itd break the test

SamHellawell (approved):

PR #93, opened 2023-01-14
Keydoc to keypair

osinakayah:


packages/core/lib/services/credential/service.js
```diff
@@ -79,11 +81,17 @@ class CredentialService {
...
-    const kp = getKeypairFromDoc(keyDoc);
-    kp.signer = kp.signer();
-    const suite = getSuiteFromKeyDoc(kp);
     vp.setHolder(keyDoc.controller);
-    return vp.sign(suite, challenge, domain);
+    keyDoc.keypair = keyDocToKeypair(keyDoc, getDock());
+    return vp.sign(
+      {
+        ...keyDoc,
+        id: `${keyDoc.controller}#keys-1`, // HACK: make it work with SDK did resolution, this is an SDK limitation
```
SamHellawell, line 89: this wont work for all keydocs, you cant assume the id in wallet's case. you have to use keydoc.id which should previously set the correct value for the DID key

packages/core/lib/services/credential/service.js
```diff
@@ -79,11 +81,17 @@ class CredentialService {
     for (const signedVC of credentials) {
       vp.addCredential(signedVC);
     }
-    const kp = getKeypairFromDoc(keyDoc);
-    kp.signer = kp.signer();
-    const suite = getSuiteFromKeyDoc(kp);
     vp.setHolder(keyDoc.controller);
-    return vp.sign(suite, challenge, domain);
+    keyDoc.keypair = keyDocToKeypair(keyDoc, getDock());
+    return vp.sign(
```
SamHellawell, line 86: why was this change required when vp.sign using suite has been working just fine afaik?

osinakayah: it does not work if the key type is Sr25519

SamHellawell (approved):

PR #97, opened 2023-02-06
cache verification status

maycon-mello:
dev

### DIFF
--- a/packages/react-native/lib/credentials/credentialHooks.js
+++ b/packages/react-native/lib/credentials/credentialHooks.js
@@ -134,11 +134,17 @@ export async function getCredentialStatus(credential) {
   }
   return CREDENTIAL_STATUS.INVALID;
 }
+
+export let cachedCredentialStatus = {};
+
 export function useGetCredentialStatus({credential}) {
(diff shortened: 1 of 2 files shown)


packages/react-native/lib/credentials/credentialHooks.js
```diff
@@ -134,11 +134,17 @@ export async function getCredentialStatus(credential) {
...
+
 export function useGetCredentialStatus({credential}) {
-  const [status, setStatus] = useState(CREDENTIAL_STATUS.PENDING);
+  const [status, setStatus] = useState(
+    cachedCredentialStatus[credential.id] || CREDENTIAL_STATUS.PENDING,
+  );
 
   useEffect(() => {
     getCredentialStatus(credential).then(response => {
+      cachedCredentialStatus[credential.id] = response;
```
mike-parkhill, line 147: do we ever refresh this?

maycon-mello: Yes, if a cached value is found it will be displayed, but still fetch the current value from network and update it if its different

mike-parkhill (approved):

PR #99, opened 2023-02-21
feat/webview sandbox

maycon-mello:
Sandbox webview for PEX filters


SamHellawell (approved):
lgtm the only thing im not so sure on is pex filtering as it requires sending all credentials into the sandbox rather than just one VP. i think we can keep it here and decide if we want that in the dock-app repo

PR #102, opened 2023-03-06
fix: issues with substrate

maycon-mello:


packages/core/lib/services/dock/service.js
```diff
@@ -68,7 +68,7 @@ export class DockService {
 
     this._setDockReady(true);
 
-    return result;
+    return true;
```
maycon-mello, line 71: It was creating a memory leak, as the sdk connection will be parsed to JSON by the rpc server.

mike-parkhill (dismissed):

mike-parkhill (approved):

PR #103, opened 2023-03-08
feat: handle didcomm messages

maycon-mello:


packages/relay-service/lib/did/did-resolver.js
```diff
@@ -0,0 +1,28 @@
...
+  DIDKeyResolver,
+  MultiResolver,
+  UniversalResolver,
+} from '@docknetwork/sdk/resolver';
+import dock from '@docknetwork/sdk';
+
+// Create a resolver in order to lookup DIDs for verifying
+export const universalResolverUrl = 'https://uniresolver.io';
+
+class WalletSDKResolver extends MultiResolver {
```
maycon-mello, line 12: Will refactor this later, need to have a DID resolver shared with all the wallet-sdk modules. Need to be moved to the `core` package. https://dock-team.atlassian.net/browse/DCKA-1502

packages/relay-service/tests/relay-service-e2e.test.js
```diff
@@ -1,9 +1,16 @@
 import {RelayService} from '../lib';
+import {dock} from '../lib/did/did-resolver';
 import {ALICE_KEY_PAIR_DOC, BOB_KEY_PAIR_DOC} from './mock-data';
 
 describe('Relay service', () => {
   const messageContent = `Test message ${Date.now()}`;
 
+  beforeAll(async () => {
+    await dock.init({
+      address: 'wss://knox-1.dock.io',
```
SamHellawell, line 10: should never connect to testnet node in tests/CI

SamHellawell: if this isnt an easy fix atm in wallet we can push it but it should use a local node always

maycon-mello: these should not run in the CI, was built for local testing

SamHellawell (approved):

PR #104, opened 2023-03-10
fix/didcomm issues

maycon-mello:


packages/relay-service/lib/index.js
```diff
@@ -35,7 +36,7 @@ const sendMessage = async ({keyPairDoc, recipientDid, message}) => {
 
   const {payload, did} = await generateSignedPayload(keyPairDoc, {
     to: recipientDid,
-    msg: JSON.stringify(jweMessage),
+    msg: toBase64(jweMessage),
```
SamHellawell, line 39: why is this change neededs? stringified json should work just fine. from the api:
```
this is incorrect, see on the API:

```
  const { payload, did } = await generateSignedPayload(keyPairDoc, {
    to: recipientDid,
    msg: typeof message === 'string' ? message : JSON.stringify(message),
  });
```

File: packages/relay-service/lib/index.js [hunk 2]
```diff
@@ -69,12 +70,14 @@ const getMessages = async ({keyPairDocs, limit = 20}) => {
...
+    const data = result.data;
+
     const messages = await Promise.all(
-      result.data.map(async item => {
+      data.map(async item => {
         const keyPairDoc = keyPairDocs.find(doc => doc.controller === item.to);
         assert(!!keyPairDoc, `keyPairDoc not found for did ${item.to}`);
         const keyAgreementKey = await getDerivedAgreementKey(keyPairDoc);
-        const jwe = JSON.parse(item.msg);
+        const jwe = fromBase64(item.msg);
```
SamHellawell, line 80: you cant assume its always base64

maycon-mello: I've updated the impl, it will now check if the value is base64 before converting.

mike-parkhill (approved):

PR #105, opened 2023-03-13
fix: did comm issues with json data

maycon-mello:


packages/relay-service/lib/index.js
```diff
@@ -80,8 +80,20 @@ const getMessages = async ({keyPairDocs, limit = 20}) => {
...
+          // TODO: check for JWT in future here
+
+          try {
+            if (isBase64(jwe)) {
+              jwe = fromBase64(jwe);
+            } else {
+              jwe = JSON.parse(jwe);
+            }
+          } catch (e) {
+            // Not valid JSON but still a string, show some kind of error?
```
maycon-mello, line 93: For now will skip this message, maybe could create a sentry error

SamHellawell: debug log could work for now, better than nothing

SamHellawell (dismissed):

packages/relay-service/lib/index.js
```diff
@@ -80,19 +81,31 @@ const getMessages = async ({keyPairDocs, limit = 20}) => {
...
+            }
+          } catch (e) {
+            Logger.debug(`Invalid JWE message received: ${jwe}`);
+            console.error(e);
+            return null;
+          }
         }
 
         const didCommMessage = await didcomm.decrypt(jwe, keyAgreementKey);
         return {
```
SamHellawell, line 100:
```
        return {
          ...item,
          msg: didCommMessage,
        };
```

is better to prevent duplication, especially for large messages

maycon-mello: fixed

SamHellawell (approved):

PR #108, opened 2023-03-29
fix: fixing issues with json-rpc on sandbox

maycon-mello:
master


packages/core/lib/sandbox.js
```diff
@@ -45,23 +34,6 @@ global.WebSocketEventTarget = () => {
   throw new Error('WebSocketEventTarget is not available in the sandbox');
 };
 
-// Disable location and redirects
```
mike-parkhill, line 48: is this functionality still being suppressed?

maycon-mello: Yes, I've added a Content-Security-Policy config in the html header.

mike-parkhill (approved):

PR #109, opened 2023-04-12
fix: issues with credential status check

maycon-mello:
Fixing a race condition bug that was exposed after making the credential screen the default tab in the app. The bug affects the credential verification check.

### DIFF
--- a/packages/react-native/lib/credentials/credentialHooks.js
+++ b/packages/react-native/lib/credentials/credentialHooks.js
@@ -2,6 +2,7 @@ import {useMemo, useCallback, useState, useEffect} from 'react';
 import {useWallet} from '../index';
 import assert from 'assert';
 import {credentialServiceRPC} from '@docknetwork/wallet-sdk-core/lib/services/credential';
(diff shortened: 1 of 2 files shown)


packages/react-native/lib/credentials/credentialHooks.js
```diff
@@ -35,8 +36,25 @@ export function getCredentialTimestamp(credential) {
...
+    setTimeout(() => {
+      clearInterval(interval);
+      reject(new Error('Timed out'));
+    }, timeout);
+  });
+}
+
 async function getCredentialValidityStatus(credential) {
   try {
+    await waitFor(() => dockService.isApiConnected(), 8000);
```
mike-parkhill, line 57: 8 seconds seems like a fairly long time out - are we hitting that regularly?

mike-parkhill (approved):

packages/react-native/lib/credentials/credentialHooks.js
maycon-mello: No, can take 3 to 4 seconds. But a bad internet connection could reach almost 8 seconds.

PR #110, opened 2023-04-19
feat/data store impl

maycon-mello:
- First PR for data store
- Created @docknetwork/wallet-sdk-data-store package
- Defined high level interface, and typeorm connection
- Defined migration process
- Will create additional PRs:
  - universal wallet import/export
  - migrating documents to new db
  - credentials table view (on top of documents) to allow better filtering/sorting
  - integrate that with the wallet-sdk-core


packages/data-store/src/configs.ts
```diff
@@ -0,0 +1,74 @@
...
+  options: DataStoreConfigs,
+): Promise<SDKConfigs> {
+  let result = await getDataSource().getRepository(SDKConfigsEntity).find();
+
+  if (result.length === 1) {
+    return result[0];
+  }
+
+  if (result.length > 1) {
+    throw new Error('More than one SDKConfigsEntity found');
```
mike-parkhill, line 15: how would the user recover from this state? I'm assuming this could potentially make the app unusable?

packages/data-store/src/migration/v2/v1-data-store.ts
```diff
@@ -0,0 +1,35 @@
...
+let _localStorage: LocalStorage;
+
+/**
+ * Legacy data-store was based on AsyncLocalStorage
+ * The wallet will inject its implementation here, so that the migration scripts can use it
+ *
+ * @param _impl
+ */
+export function setV1LocalStorage(_impl: LocalStorage) {
+  _localStorage = _impl;
```
mike-parkhill, line 16: could `_impl` be null or undefined here?

packages/data-store/src/typeorm/entities/document.entity.ts
```diff
@@ -0,0 +1,121 @@
...
+  })
+  @JoinTable()
+  _typeRel: Relation<DocumentTypeEntity[]>;
+
+  @Column('blob')
+  data: string;
+}
+
+export async function getOrCreateDocumentTypes(
+  types,
```
mike-parkhill, line 43: cam we enforce a type here?

packages/data-store/src/typeorm/entities/document.entity.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  return typeEntityList;
+}
+
+export async function createDocument(json: any): Promise<DocumentEntity> {
+  assert(!!json.id, 'Document must have an id');
+
+  // TODO: get walletId from context
+  const walletId = await getActiveWallet();
+
```
mike-parkhill, line 79: when are these TODOs getting handled?

packages/data-store/src/typeorm/entities/document.entity.ts
```diff
@@ -0,0 +1,121 @@
...
+  };
+
+  const repository = getDataSource().getRepository(DocumentEntity);
+
+  return repository.save(entity);
+}
+
+export async function getDocumentById(id: string): Promise<DocumentEntity> {
+  const repository = getDataSource().getRepository(DocumentEntity);
+  const walletId = await getActiveWallet();
```
mike-parkhill, line 103: should the walletId be coming from context here too?

packages/data-store/src/typeorm/entities/document.entity.ts
```diff
@@ -0,0 +1,121 @@
...
+      walletId,
+    },
+  });
+}
+
+export async function getDocumentsByType(
+  type: string,
+): Promise<DocumentEntity[]> {
+  const repository = getDataSource().getRepository(DocumentEntity);
+  const walletId = await getActiveWallet();
```
mike-parkhill, line 117: and here

packages/data-store/src/typeorm/index.ts
```diff
@@ -0,0 +1,47 @@
...
+import {WalletEntity} from './entities/wallet.entity';
+import {DocumentEntity} from './entities/document.entity';
+import {DocumentTypeEntity} from './entities/document-type.entity';
+
+let _dataSource: DataSource;
+
+export async function initializeTypeORM(options: DataStoreConfigs) {
+  let databasePath = 'wallet-data-store.sqlite';
+
+  if (process.env.NODE_ENV === 'test') {
```
mike-parkhill, line 15: bit of a nitpick, but rather than expecting the datastore to know that tests are a thing at all, why not make DATASTORE_PATH an env var that defaults to `wallet-data-store.sqlite` but tests can set to `:memory:`?

packages/data-store/src/typeorm/index.ts
```diff
@@ -0,0 +1,47 @@
...
+      WalletEntity,
+      DocumentEntity,
+      DocumentTypeEntity,
+    ],
+    // TODO: will remove this once we have all entities in place
+    synchronize: true,
+    dropSchema: options.dropSchema,
+  });
+
+  await _dataSource
```
mike-parkhill, line 34: how does closing the datastore get handled? If the app is crashing is there something we can hook into to capture and try to shutdown gracefully?

packages/data-store/src/configs.ts
maycon-mello: Maybe we can return only the first item, and log a warning on sentry, allowing us to track this state without braking the app.

packages/data-store/src/typeorm/entities/document.entity.ts
maycon-mello: Actually this looks correct, just forgot to remove the TODO comment. The wallet id was hardcoded when I added the TODO

packages/data-store/src/typeorm/entities/document.entity.ts
maycon-mello: fixed

packages/data-store/src/migration/v2/v1-data-store.ts
maycon-mello: fixed

packages/data-store/src/migration/index.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const migrations = [
+  migrateToV2,
+  // add new migrations here
+];
+
+export async function migrate(options: DataStoreConfigs) {
+  let sdkConfigs: SDKConfigs = await getSDKConfigs(options);
+
+  for (const migrate of migrations) {
```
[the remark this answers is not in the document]
maycon-mello: Created a ticket for it https://dock-team.atlassian.net/browse/DCKA-1623

packages/data-store/src/configs.ts
maycon-mello: Created a follow up sub task https://dock-team.atlassian.net/browse/DCKA-1624

packages/data-store/src/typeorm/index.ts
maycon-mello: I've added a function to close the connection. Also created a jira to follow up that on the wallet app side. https://dock-team.atlassian.net/browse/DCKA-1625.

The app will need to detect an exist/crash event and call the wallet-sdk function

packages/data-store/src/typeorm/index.ts
maycon-mello: yeah, much better. I've pushed a commit for that

packages/data-store/src/typeorm/entities/document.entity.ts
maycon-mello: fixed, I've removed the TODO comment

packages/data-store/src/typeorm/index.ts
mike-parkhill: do we need to worry about transaction support so we can do rollbacks or do you think none of our scenarios will need it?

integration-tests/accounts.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  });
+
+  it('expect to import account from JSON', async () => {
+    let account = await getWallet().getDocumentById(AccountJSON.address);
+    expect(account).toBeUndefined();
+
+    await importAccountJSON();
+    account = await getWallet().getDocumentById(AccountJSON.address);
+
+    console.log(account);
```
mike-parkhill, line 63: is this log still needed?

integration-tests/wallet-snapshot-v1.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    const [account] = accounts;
+
+    expect(account.address).toBe(
+      '37NKEP14n9omsAgxJ3sn14XtHo2vs5a34UcpCyybXQJoUQ92',
+    );
+
+    await assertAccountIsValid(account.address);
+  });
+
+  it('expect to have loaded DIDs', async () => {});
```
mike-parkhill, line 41: what's the plan for these empty tests?

packages/data-store/src/index.ts
```diff
@@ -26,3 +26,11 @@ export async function createDataStore(_options: DataStoreConfigs = {}) {
 
   return true;
 }
+
+/**
+ * Close the data store connection with the database
+ *
+ */
+export async function closeDataStore() {
+  return getDataSource().destroy();
```
mike-parkhill, line 35: do we need to explicitly issue a final commit before closing it?

packages/data-store/src/configs.ts
mike-parkhill: We should probably try to add some safety during insert to ensure we aren't allowing a second one to be added in the first place.

integration-tests/wallet-snapshot-v1.test.ts
maycon-mello: I have another branch with those, wanted to push as separate PR to make it easier to review.

packages/data-store/src/index.ts
maycon-mello: No that's not required

integration-tests/wallet-snapshot-v1.test.ts
maycon-mello: follow up PR https://github.com/docknetwork/react-native-sdk/pull/111

integration-tests/accounts.test.ts
maycon-mello: Log was removed here https://github.com/docknetwork/react-native-sdk/pull/111

mike-parkhill (approved):

PR #111, opened 2023-04-21
feat/wallet integration tests

maycon-mello:
- integration tests


integration-tests/data/verification-templates/any-basic-credential.json
mike-parkhill, line 1: are these templates meant to be empty?

maycon-mello: yes, I didn't create the integration tests for those yet. Will push as separate PR

PR #112, opened 2023-04-26
moving webview code to separate package

maycon-mello:


mike-parkhill:
are there any real changes? Or is it just moving and updating references?

maycon-mello:
> are there any real changes? Or is it just moving and updating references?

No real changes, only moving and renaming the package name

PR #113, opened 2023-04-27
feat: resolve DIDComm messages

maycon-mello:
Methods to decrypt DIDComm messages, will be integrated with the qr-code handler in the app side.

### DIFF
--- a/packages/relay-service/lib/index.js
+++ b/packages/relay-service/lib/index.js
@@ -75,30 +75,14 @@ const getMessages = async ({keyPairDocs, limit = 20}) => {
     const data = result.data;
 
     const messages = await Promise.all(
-      data.map(async item => {
-        const keyPairDoc = keyPairDocs.find(doc => doc.controller === item.to);
-        assert(!!keyPairDoc, `keyPairDoc not found for did ${item.to}`);
-        const keyAgreementKey = await getDerivedAgreementKey(keyPairDoc);
-        let jwe = item.msg;
+      data.map(async message => {
+        const didCommMessage = await resolveDidcommMessage({
+          message,
+          keyPairDocs,
+        });
 
-        if (typeof jwe === 'string') {
-          // TODO: check for JWT in future here
-          try {
-            if (isBase64(jwe)) {
-              jwe = fromBase64(jwe);
-            } else {
-              jwe = JSON.parse(jwe);
-            }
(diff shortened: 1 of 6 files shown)


packages/relay-service/tests/relay-service.test.js
```diff
@@ -93,4 +93,51 @@ describe('Relay service', () => {
...
+          msg: toBase64(didCommMessage),
+        },
+        keyPairDocs: [BOB_KEY_PAIR_DOC],
+      });
+
+      expect(result.payload).toStrictEqual({test: 'test'});
+    });
+
+    it('expect to handle URL', async () => {
+      // TODO: mock axios.get to return didComm message
```
mike-parkhill, line 124: when are we coming back for this?

maycon-mello: sorry, it was work in progress. I've pushed the final commit now

maycon-mello: added the axios mock, and the ability to resolve messages from URLs

mike-parkhill (dismissed):

SamHellawell (approved):

maycon-mello:
Pushed new commit to fix linting, + improvement on recipient resolution

packages/relay-service/lib/index.js
```diff
@@ -200,11 +205,20 @@ export async function resolveDidcommMessage({keyPairDocs, message}) {
 
   let result = jwe;
 
-  if (message.to) {
-    const keyPairDoc = keyPairDocs.find(doc => doc.controller === message.to);
+  let recipients = jwe?.recipients?.map(recipient => recipient.header.kid);
+
+  if (recipients) {
```
mike-parkhill, line 210: is it valid to have a message with no recipients?

maycon-mello: if it's not a didcomm message, yes. Maybe I should rename this variables, its being used to check if didcomm.decrypt must be used or not

maycon-mello: let me update that

maycon-mello: done, i've pushed an update. This case applies for jwt message, which is not encrypted using didcomm, but should be a valid message for us

mike-parkhill (approved):

PR #115, opened 2023-05-03
updated wallet-core, network resolver

maycon-mello:
- wallet core methods
- splitting document entity methods into multiple files, for better maintenance
- impl for document network resolver
- improved data migration
- unit tests


integration-tests/switch-wallet.test.ts
```diff
@@ -2,20 +2,57 @@ import {
...
-    await createAccounts();
   });
 
-  it('expect to import data in the mainnet wallet', () => {});
+  it('expect to filter mainnet documents from testnet documents', async () => {
+    await setNetwork('mainnet');
+    await createWalletData();
+    await assertWalletData();
+    await setNetwork('testnet');
+    // const isEmpty = await isWalletEmpty();
```
mike-parkhill, line 44: what's happening with this commented code?

packages/core/src/network-resolver.test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  it('expect to add document to testnet', async () => {
+    const mockDocument = {
+      id: 'test',
+      type: 'VerifiableCredential',
+    };
+
+    wallet.setNetworkId('testnet');
+
+    await wallet.addDocument(mockDocument);
```
mike-parkhill, line 35: is it worth asserting that the document _wasn't_ added to the other network?

packages/core/src/network-resolver.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  return null;
+}
+
+async function didResolver({
+  document,
+  dataStore,
+}: DocumentResolverProps): Promise<ResolverResult> {
+  return null;
```
mike-parkhill, line 59: is this still pending?

packages/core/src/network-resolver.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  if (document.id.indexOf('creds-testnet.dock.io') > -1) {
+    return 'testnet';
+  }
+
+  if (document.id.indexOf('creds.dock.io') > -1) {
+    return 'mainnet';
+  }
+
```
mike-parkhill, line 51: do we need to fall back to a DID resolution check here?

packages/data-store/src/configs.ts
```diff
@@ -1,74 +1,27 @@
...
+export const DEFAULT_CONFIGS: DataStoreConfigs = {
+  defaultNetwork: 'mainnet',
+  nonIsolatedNetworks: false,
+  databasePath: 'dock-wallet-sdk.db',
+  documentNetworkResolver: genericDocumentNetworkResolver,
   networks: [
     {
       name: 'Mainnet',
       id: 'mainnet',
+      configs: {
```
mike-parkhill, line 13: how would we inject new networks?

packages/data-store/src/configs.ts
```diff
@@ -1,74 +1,27 @@
...
-export const defaultDataStoreConfigs: DataStoreConfigs = {
-  sdkConfigs: {
-    activeWallet: 'mainnet',
-  },
+import {DataStoreConfigs} from './types';
+import {genericDocumentNetworkResolver} from './document-network-resolver';
+
+export const DEFAULT_CONFIGS: DataStoreConfigs = {
+  defaultNetwork: 'mainnet',
+  nonIsolatedNetworks: false,
```
mike-parkhill, line 6: what is this flag used for?

packages/data-store/src/configs.ts
```diff
@@ -1,74 +1,27 @@
...
-
-export const defaultDataStoreConfigs: DataStoreConfigs = {
-  sdkConfigs: {
-    activeWallet: 'mainnet',
-  },
+import {DataStoreConfigs} from './types';
+import {genericDocumentNetworkResolver} from './document-network-resolver';
+
+export const DEFAULT_CONFIGS: DataStoreConfigs = {
+  defaultNetwork: 'mainnet',
```
mike-parkhill, line 5: if we had to support a completely different set of networks (e..g in a white label build for a private version of Dock) how would we override the defaults if they're hard-coded here?

packages/data-store/src/entities/document/document.entity.ts
```diff
@@ -0,0 +1,41 @@
...
+  })
+  @JoinTable()
+  _typeRel: Relation<DocumentTypeEntity[]>;
+
+  // Add the ManyToMany self-referencing relationship to track document correlations
+  @ManyToMany(() => DocumentEntity, {cascade: true, nullable: true})
+  @JoinTable()
+  correlation: DocumentEntity[];
+
+  // TODO: add correlations
```
mike-parkhill, line 37: when is this to happen?

mike-parkhill, line 37: when is this to happen?

maycon-mello: I've removed the TODO. This is already implemented.

packages/data-store/src/entities/document/create-document.ts
```diff
@@ -0,0 +1,54 @@
...
+ * @param dataStore
+ * @param json
+ */
+export async function createDocument({
+  dataStore,
+  json,
+}: ContextProps & {
+  json: any;
+}): Promise<WalletDocument> {
+  assert(!!json.id, 'Document must have an id');
```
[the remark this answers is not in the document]
maycon-mello: I've removed this validation and will also add some tests for JWT-VC

packages/data-store/src/configs.ts
maycon-mello: No, I've created an example of how to do that on this commit: https://github.com/docknetwork/react-native-sdk/pull/115/commits/3e99e343b144e0eb17d7e9db1dd43fba748d184b

Added a new integration test to cover custom networks and JWT-VC

packages/data-store/src/configs.ts
maycon-mello: Let me remove this flag, we are not using that. I was thinking about the possibility to allow handling documents from multiple networks in one wallet, the same approach we have now, the queries would not be scoped by the network, but for now, that's not a use case

packages/data-store/src/configs.ts
maycon-mello: I've created an example here https://github.com/docknetwork/react-native-sdk/pull/115/commits/3e99e343b144e0eb17d7e9db1dd43fba748d184b

integration-tests/switch-wallet.test.ts
maycon-mello: fixed, should be running now

packages/core/src/network-resolver.test.ts
maycon-mello: sure, added it here https://github.com/docknetwork/react-native-sdk/pull/115/commits/eda903201840da84eefb332862055a3d49fcd897

packages/core/src/network-resolver.ts
maycon-mello: yes, I will add another PR with this logic

maycon-mello: Still testing with the dock-app, want to define the better logic for it yet

packages/core/src/network-resolver.ts
maycon-mello: I think so; I've added a TODO for it and will implement it along with the account resolver

integration-tests/custom-networks.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    },
+    {
+      name: 'Polygon mumbai',
+      id: 'mumbai',
+      configs: '{}',
+    },
+  ];
+
+  it('should create a wallet with custom networks', async () => {
+    wallet = await createWallet({
```
mike-parkhill, line 19: is there anything we could assert about the wallet to make sure it was created properly?

integration-tests/custom-networks.test.ts
```diff
@@ -0,0 +1,55 @@
...
+      databasePath: ':memory:',
+      networks,
+    });
+  });
+
+  it('should add documents to polygon network', async () => {
+    await wallet.addDocument({
```
mike-parkhill, line 26: how do we know the document was added? This proves the call didn't blow up, but it doesn't assert that the doc was stored properly

integration-tests/custom-networks.test.ts
```diff
@@ -0,0 +1,55 @@
...
+      type: 'JWT-VC',
+    });
+
+    documents = await wallet.getAllDocuments();
+
+    expect(documents).toHaveLength(1);
+  });
+
+  it('should switch back to polygon network and have a document', async () => {
+    wallet.setNetworkId('polygon');
```
mike-parkhill, line 50: there's a danger here that if anyone ever re-orders the tests that this will fail since it depends on the previous tests

integration-tests/custom-networks.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    const documents = await wallet.getAllDocuments();
+
+    expect(documents).toHaveLength(1);
```
mike-parkhill, line 53: since both mumbai and polygon each have a single document at this point this isn't proving that it's the right document (or that the network actually switched)

packages/core/src/network-resolver.ts
mike-parkhill: should we allow a mapping between hostname and network to be injected with the network configs? Just thinking of how we would do this for other networks. Maybe have the network definition be something like

```
    {
      name: 'polygon',
      id: 'polygon',
      credential-urls: [],
      configs: '{}',
    }
```

and then this resolver would just look for a match in the array of linked credential-urls

integration-tests/custom-networks.test.ts
maycon-mello: Yes, I've updated the unit test. The wallet will be created in a beforeEach function. An empty wallet for each test

integration-tests/custom-networks.test.ts
maycon-mello: I've updated the test

integration-tests/custom-networks.test.ts
maycon-mello: yeah, I've updated that, using a beforeEach function which will create a new wallet for each test

integration-tests/custom-networks.test.ts
maycon-mello: yeah, I've created two different mock documents now. So that we can identify that by using the document value

packages/core/src/network-resolver.ts
maycon-mello: that's a good idea, looks very flexible. Let me add this to this PR

mike-parkhill: Ideally we won't have 'mainnet' or 'dock.io' or anything like that hard-coded in either app when we're done

mike-parkhill (approved):

PR #116, opened 2023-05-18
Fix isDockCertsURL to exclude /proof/

SamHellawell:


packages/credentials/lib/index.js
```diff
@@ -77,9 +77,11 @@ export class Credentials {
    */
   isDockCertsURL(url) {
```
mike-parkhill, line 78: we might want to rename this to `isDockCertsCredentialUrl` or something

mike-parkhill: doesn't have to be now, but...

SamHellawell: was thinking same

mike-parkhill (approved):

mike-parkhill (approved):

PR #117, opened 2023-05-24
fixing issues with wallet-app integration

maycon-mello:


packages/react-native/lib/index.js
```diff
@@ -92,42 +96,46 @@ export function useAccount(address) {
...
-  const sandboxWebViewRef = useRef();
-  const baseUrl =
-    Platform.OS === 'ios' ? 'app-html' : 'file:///android_asset/app-html';
+  const [firstFetch, setFirstFetch] = useState();
+  const networkId = useMemo(() => {
+    return wallet?.getNetworkId();
+  }, [wallet]);
+
+  console.log('Current networkId', networkId);
+  console.log('wallet documents', documents);
```
mike-parkhill, line 114: we probably don't want to dump all the documents out - that could be pretty large, right? and possibly hold PII

packages/react-native/lib/wallet.ts
```diff
@@ -0,0 +1,29 @@
...
+  if (!wallet) {
+    wallet = await initializeWallet();
+  }
+
+  return wallet;
+}
+export async function initializeWallet() {
+  wallet = await createWallet({
+    databasePath: 'dock-wallet',
+    dbType: 'react-native',
```
mike-parkhill, line 24: I wouldn't expect a `dbType` to be `react-native`. I was expecting a database related value like postgres or mysql or sqlite or something

maycon-mello: I agree, but this is the constant value defined on `typeorm` side. https://github.com/typeorm/react-native-example/blob/master/src/App.tsx#L42

`react-native` is used for mobile a sqlite, and is mentioned on their docs. For nodejs `sqlite` should be used. We could map it out on our side, but could create breaking changes and not be consistent with `typeorm` docs

mike-parkhill: ah, that's an odd choice on their part

no need to complicate things on our side

packages/react-native/lib/index.js
maycon-mello: Fixed

packages/wasm/lib/services/credential/service.js
```diff
@@ -15,6 +15,7 @@ import {verifyCredential} from '@docknetwork/sdk/utils/vc/credentials';
 import {PEX} from '@sphereon/pex';
 import {keyDocToKeypair} from './utils';
 import {getDock} from '../dock/service';
+import {Logger} from '../../core/logger';
```
mike-parkhill, line 18: lint

mike-parkhill (approved):

mike-parkhill:
Still has a lint/test failure but otherwise is good

integration-tests/switch-wallet.test.ts
```diff
@@ -36,14 +36,14 @@ describe('Switch wallet', () => {
...
 
-  it('expect to filter mainnet documents from testnet documents', async () => {
-    await setNetwork('mainnet');
-    await createWalletData();
-    await assertWalletData();
-    await setNetwork('testnet');
-    const isEmpty = await isWalletEmpty();
-    expect(isEmpty).toBe(true);
-  });
+  // it('expect to filter mainnet documents from testnet documents', async () => {
```
mike-parkhill, line 39: is there a plan for restoring this test?

maycon-mello: Yes, https://dock-team.atlassian.net/browse/DCKA-1708. I will fix those today

maycon-mello: I will push another PR to fix that

mike-parkhill (approved):

PR #135, opened 2023-07-18
Feat/build modules

maycon-mello:
- Moving ts/js files to `src` folder
- Will be using `lib` folder for transpiled code
- created example for nodejs, working fine
- remove direct `typeorm` imports on data-store package, switching to a proxy file which exposes `typeorm`. So that we can override it with `typeorm/web` more easily for web builds.
- working on web example (working in progress)


packages/wasm/src/services/pex/service-rpc.js
```diff
@@ -12,8 +12,9 @@ export class PEXServiceRPC extends RpcService {
     this.sandbox = process.env.NODE_ENV !== 'test';
   }
   async filterCredentials(params: FilterCredentialsParams) {
-    validation.filterCredentials(params);
-    return this.call('filterCredentials', params);
+    // validation.filterCredentials(params);
+    // return this.call('filterCredentials', params);
+    alert('filter credential......');
```
maycon-mello, line 17: testing code, needs to be removed

mike-parkhill:
206 files!

How much is new and how much is moved? i.e. can skip some stuff?

maycon-mello:
> 206 files!
> 
> How much is new and how much is moved? i.e. can skip some stuff?
The biggest change was on this file packages/wasm/src/services/dock/service.js.

Other changes were basically moving files from `lib` folder to `src` folder and adding a build process for all the packages. It should generate the `lib` folder with plain js esm modules, which can be imported on web, and node.js.

packages/data-store/src/types.ts
```diff
@@ -44,7 +44,7 @@ export type DataStoreConfigs = {
 
   // Typeorm supported multiple RDBMS types https://typeorm.io/data-source-options#common-data-source-options
   // We only tested sqlite and react-native sqlite so far
```
mike-parkhill, line 46: should we update this comment? Or is sqljs just another sqlite wrapper?

maycon-mello: good point, sqljs should be added

maycon-mello: let me push an update here

packages/wasm/src/core/realm.js
```diff
@@ -1,10 +1,21 @@
...
 import {Account, TokenPrice, RequestLog} from './realm-schemas';
 
+let Realm;
 let realm;
 
 const schema = [TokenPrice, Account, RequestLog];
 
+/**
+ *
+ * Realm is being used by the dock-app
```
mike-parkhill, line 10: if Realm is only intended to be used by the dock app should we move the code? Or is the move to typeorm soon enough that it's not worth it?

maycon-mello: We should move to typeorm soon enough. That's not a big lift, so could be handled in the next month. https://dock-team.atlassian.net/browse/DCKA-1699

mike-parkhill (approved):

PR #137, opened 2023-07-27
wallet-sdk cli

maycon-mello:


packages/cli/src/commands/dids.ts
```diff
@@ -0,0 +1,29 @@
...
+    console.log(dids);
+  });
+
+didsCommand
+  .command('create')
+  .description('List wallet dids')
+  .action(async () => {
+    const wallet: IWallet = await getWallet();
+
+    // implement did creation logic, start with did key
```
mike-parkhill, line 26: are you tracking all these improvement thoughts anywhere else? I feel like these comments might get lost

mike-parkhill (approved):
I think this will be very handy for dev and test

maycon-mello:
> I think this will be very handy for dev and test

Yeah, I was planning to publish it as a built-in package (all the deps included in one single file). So it should have zero setup time.

It should work like `npx @docknetwork/wallet-sdk-cli credentials list`. Only requirement is to have npm and nodejs

packages/cli/src/commands/dids.ts
maycon-mello: Yes. I pushed this as placeholder for only. But I'm currently working on migrating the basic verification flow to the wallet-sdk. DID creation is a requirements on it. And I will leave wallet-backups (import, export), accounts, and transactions for the future.

PR #138, opened 2023-08-01
fixing issues with android data-store

maycon-mello:


mike-parkhill (approved):
Is there an e2e test we can add to the list that @Jayteekay is writing to cover this?

maycon-mello:
II think it is worth adding this as an e2e test scenario to ensure the wallet can handle that

mike-parkhill (approved):

PR #144, opened 2023-08-16
Feat/verification provider + range proofs on wallet-sdk-cli

maycon-mello:
- The current verification flow was partially implemented in the wallet-sdk and some logic was placed in the wallet-app. We had a lot of hooks to, which were hard to test. I've moved all its logic to the verificationProvider, so that we have a single place to implement the verification logic and it will be better to maintain


packages/cli/src/commands/messages.ts
```diff
@@ -12,44 +14,71 @@ messagesCommand
...
     });
 
-    // send relay service message
-    // if multiple dids, ask with did to use
+    console.log('Message sent!');
+    console.log(result);
+  });
+
+const DIDAddressBook = {
+  iosEmulator: 'did:key:z6MkuZRbStWnxNunsGRbWBfMdZaE9iqrgFkYZXi1C6dMtREi',
```
mike-parkhill, line 55: should these be read in from a config or env file?

packages/cli/src/commands/verification.ts
```diff
@@ -1,28 +1,117 @@
...
-import axios from 'axios';
+import select, {Separator} from '@inquirer/select';
+import checkbox from '@inquirer/checkbox';
+import input from '@inquirer/input';
 import {IWallet} from '@docknetwork/wallet-sdk-core/lib/types';
 import {getWallet} from '../helpers';
-
-const verificationCommands = new Command('dids');
+import {createVerificationController} from '@docknetwork/wallet-sdk-core/src/verification-controller';
+// import template from '@docknetwork/wallet-sdk-core/src/fixtures/iiw-template.json';
```
mike-parkhill, line 8: is this commented import needed?

packages/core/src/credential-provider.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+export function createCredentialProvider({wallet}: {wallet: IWallet}) {
+  function getCredentials(type: string = 'VerifiableCredential') {
+    return wallet.getDocumentsByType(type);
+  }
+
+  return {
+    getCredentials,
+    // TODO: move credential validity check to this provider
```
mike-parkhill, line 14: are these still in the app codebase?

packages/wasm/src/services/credential/service.ts
```diff
@@ -167,6 +176,8 @@ class CredentialService {
         }
       });
       credentialJSON['@context'].push(customContext);
+
+      console.log();
```
mike-parkhill, line 180: ?

maycon-mello: I've removed it

packages/cli/src/commands/verification.ts
maycon-mello: I've removed it

packages/core/src/credential-provider.ts
maycon-mello: Yes, still need to move this part

packages/cli/src/helpers.ts
```diff
@@ -25,3 +32,29 @@ export async function getWallet(): Promise<IWallet> {
 export function getDIDProvider(): IDIDProvider {
   return didProvider;
 }
+
+export function getCredentialProvider(): ICredentialProvider {
+  return credentialProvider;
+}
+
+export async function selectCredential() {
+  const credentials = await credentialProvider.getCredentials();
```
mike-parkhill, line 41: would be cool if the credentialProvider could just iterate over a folder so I could throw different ones easily as I'm testing

mike-parkhill (approved):

packages/react-native/bundler/index.js
```diff
@@ -44,13 +45,34 @@ const getWebpackConfig = ({entry, path, filename}) => ({
   },
   module: {
     rules: [
+      // {
```
mike-parkhill, line 48: are these commented sections coming back?

mike-parkhill (approved):

PR #145, opened 2023-08-21
fix react-native data migration

maycon-mello:
rewriting the migration scripts to use TypeORM helpers instead of plain SQL, it should provide better compatibility across different DBs and devices

### DIFF
--- a/packages/data-store/src/migrations/1691498362273-bootstrap.ts
+++ b/packages/data-store/src/migrations/1691498362273-bootstrap.ts
@@ -1,94 +1,125 @@
-import {MigrationInterface, QueryRunner} from 'typeorm';
+import {
+  MigrationInterface,
+  QueryRunner,
+  Table,
+  TableColumn,
+  TableForeignKey,
+  TableIndex,
+} from 'typeorm';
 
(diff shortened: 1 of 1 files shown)


packages/data-store/src/migrations/1691498362273-bootstrap.ts
```diff
@@ -1,94 +1,125 @@
...
-    await queryRunner.query(
-      `CREATE TABLE "wallet_entity" ("id" varchar PRIMARY KEY NOT NULL, "version" varchar, "networkId" varchar NOT NULL)`,
-    );
-    await queryRunner.query(
-      `CREATE TABLE "document_entity__type_rel_document_type_entity" ("documentEntityId" text NOT NULL, "documentTypeEntityId" text NOT NULL, PRIMARY KEY ("documentEntityId", "documentTypeEntityId"))`,
+    await queryRunner.createTable(
+      new Table({
+        name: 'network_entity',
+        columns: [
+          new TableColumn({name: 'id', type: 'varchar', isPrimary: true}),
```
mike-parkhill, line 27: is "NOT NULL" implied with the ORM whenever you mark a primary index?

PR #150, opened 2023-09-08
feat/bound checks

maycon-mello:
range proofs bound checks

depends on https://github.com/docknetwork/web/pull/394

### DIFF
--- a/packages/wasm/src/services/credential/bound-check.test.ts
+++ b/packages/wasm/src/services/credential/bound-check.test.ts
@@ -206,4 +206,59 @@ describe('Bound check', () => {
       provingKey,
     );
   });
+
+  it('expect to create bound check for minimum and maximum number', () => {
+    applyEnforceBounds({
+      builder,
+      proofRequest: createProofRequest([
+        {
+          path: ['$.credentialSubject.income'],
+          filter: {
+            type: 'number',
+            formatMaximum: 20000,
+            formatMinimum: 5000,
+          },
+          predicate: 'required',
+        },
+      ]),
+      provingKey,
+      provingKeyId,
+    });
+
+    const credIdx = 0;
+    const attributeName = 'credentialSubject.income';
+    const min = 5000;
+    const max = 20000;
+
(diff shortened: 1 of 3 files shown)


packages/wasm/src/services/credential/bound-check.ts
```diff
@@ -0,0 +1,127 @@
...
+
+export async function fetchProvingKey(proofRequest: ProofRequest) {
+  let provingKey: LegoProvingKey;
+  let blob: Uint8Array;
+
+  if (isBase64(proofRequest.boundCheckSnarkKey)) {
+    const base64Data = proofRequest.boundCheckSnarkKey;
+    blob = Buffer.from(base64Data, 'base64');
+  } else {
+    const response = await fetch(proofRequest.boundCheckSnarkKey);
```
maycon-mello, line 113: I'm currently working on caching this data. Will push an additional PR for it

packages/wasm/src/services/credential/bound-check.test.ts
```diff
@@ -0,0 +1,209 @@
...
+      credIdx,
+      attributeName,
+      min,
+      max,
+      provingKeyId,
+      provingKey,
+    );
+  });
+
+  it('expect to create bound check for minimum number', () => {
```
mike-parkhill, line 178: we should have a test for numbers with max and minimum too, I think

packages/wasm/src/services/credential/bound-check.ts
```diff
@@ -0,0 +1,127 @@
...
+
+      if (format === 'date-time') {
+        max = formatMaximum
+          ? dateTimeToTimestamp(formatMaximum)
+          : MAX_DATE_PLACEHOLDER;
+        min = formatMinimum ? dateTimeToTimestamp(formatMinimum) : 0;
+      } else if (type === 'number') {
+        max = formatMaximum || MAX_NUMBER;
+        min = formatMinimum || 0;
+      }
```
mike-parkhill, line 83: if it's an unknown format, should we error out?

mike-parkhill (approved):

PR #151, opened 2023-09-11
fixing build issue

maycon-mello:


mike-parkhill (approved):
Shouldn't linting have caught this?

maycon-mello:
> Shouldn't linting have caught this?

The rollup build was failing because it was not able to parse typescript. I have another PR adding support for typescript on it, It also adds the rollup build as part of the CI so that we can catch these build issues too

PR #154, opened 2023-09-15
create default did on testnet

maycon-mello:
master

### DIFF
--- a/packages/core/src/wallet.test.ts
+++ b/packages/core/src/wallet.test.ts
@@ -1,3 +1,4 @@
+import {createDIDProvider} from './did-provider';
 import {createWallet, ensureDocumentContext, IWallet} from './wallet';
 
 describe('Wallet', () => {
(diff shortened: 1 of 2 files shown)


packages/wasm/src/services/credential/bound-check.test.ts
```diff
@@ -261,32 +261,4 @@ describe('Bound check', () => {
       'Unsupported format undefined and type string for enforce bounds',
     );
   });
-
```
mike-parkhill, line 264: ?

maycon-mello: this commit is from this PR https://github.com/docknetwork/react-native-sdk/pull/155. We are dropping the dateTimeToTimestamp function as the crypto-wasm-ts will implement it. 

let me rebase this PR

maycon-mello: it's done, the branch is rebased

packages/core/src/wallet.test.ts
```diff
@@ -31,6 +32,15 @@ describe('Wallet', () => {
     expect(result[1]['@context']).toEqual('test');
   });
 
+  it('expect to create default dids', async () => {
+    const dids = await createDIDProvider({wallet}).getAll();
+    expect(dids.length).toBe(1);
+    await wallet.setNetwork('testnet');
+    await new Promise(resolve => setTimeout(resolve, 1000));
+    const dids2 = await createDIDProvider({wallet}).getAll();
+    expect(dids2.length).toBe(1);
```
mike-parkhill, line 41: can we assert that `dids[0] != dids2[0]`?

also minor thing, but could we call those variables `mainnetDIDs` and `testnetDIDs` to be explicit?

PR #163, opened 2023-10-03
Feat/refactor hooks

Jayteekay:
### Proposed changes

- Moving DID operations from didManagement hook to didProvider
- Refactoring/removing unnecessary codes


maycon-mello (approved):
looks good

PR #166, opened 2023-10-06
DCKA-2041: persists didcomm encrypted messages in the wallet

maycon-mello:
main


packages/core/src/message-provider.ts
```diff
@@ -0,0 +1,102 @@
...
+        }
+        await relayService.sendMessage({ keyPairDoc, message, recipientDid });
+      } catch (error) {
+        throw new Error(`Failed to send message: ${error.message}`);
+      }
+    },
+
+    async processDIDCommMessages() {
+      try {
+        const messages = await wallet.getDocumentsByType('DIDCommMessage');
```
mike-parkhill, line 50: should `DIDCommMessage` be a shared constant somewhere?

packages/core/src/message-provider.ts
```diff
@@ -0,0 +1,102 @@
...
+      try {
+        const keyPairDocs = await getKeyPairDocs(didProvider);
+        const encryptedMessages = await relayService.getMessages({
+          keyPairDocs,
+          limit: FETCH_MESSAGE_LIMIT,
+          skipMessageResolution: true,
+        });
+
+        for (const { _id, encryptedMessage } of encryptedMessages) {
+          await wallet.addDocument({
```
mike-parkhill, line 85: should the try/catch be inside the loop so we keep processing other messages? Or are you just going with giving up on this batch and then we'll refetch them all?

(can this addDocument() call even fail?)

(do we need to consider any scenarios where a bad message of some sort repeatedly fails?)

maycon-mello: I've added better logging and error handlers. In case it fails we will have proper sentry logs here

packages/core/src/helpers.ts
```diff
@@ -1,8 +1,13 @@
 import axios from 'axios';
 import assert from 'assert';
 
-let sentryCaptureException: any = (error) => {
-  console.error(error); 
+let sentryCaptureException: any = error => {
+  console.error(error);
```
cykoder, line 5: i know its not changed in this PR, but should this be submitting to sentry?

packages/core/src/message-provider.ts
```diff
@@ -91,11 +91,17 @@ export function createMessageProvider({
         });
 
         for (const { _id, encryptedMessage } of encryptedMessages) {
-          await wallet.addDocument({
-            id: _id,
-            type: 'DIDCommMessage',
-            encryptedMessage,
-          });
+          try {
+            await wallet.addDocument({
```
cykoder, line 95: at what point are these messages removed?

packages/core/src/message-provider.ts
```diff
@@ -91,11 +91,17 @@ export function createMessageProvider({
...
+              id: _id,
+              type: WalletDocumentTypes.DIDCommMessage,
+              encryptedMessage,
+            });
+          } catch(err) {
+            // this message will be lost if it fails to be stored in the wallet
+            captureException(err);
+            console.error(err);
```
cykoder, line 103: there shouldnt be a need to call console.error when you invoke `captureException` - cant `captureException` also do the console log?

packages/core/src/helpers.ts
maycon-mello: the wallet app will inejct the react-native sentry implementation

maycon-mello: let me add a comment on this line

packages/core/src/message-provider.ts
maycon-mello: there is a markMessageAsRead function exposed on this file. The wallet will listen for a `wallet.eventManager.emit('didcomm-message-decrypted', decryptedMessage);` and will call markMessageAsRead after processing it

packages/core/src/message-provider.ts
maycon-mello: that's right. I've removed it

packages/core/src/helpers.ts
maycon-mello: it should be updated now

cykoder (approved):

PR #167, opened 2023-10-09
fixing range proofs bound check

maycon-mello:
main

### DIFF
--- a/packages/wasm/src/services/credential/bound-check.ts
+++ b/packages/wasm/src/services/credential/bound-check.ts
@@ -74,26 +74,23 @@ export function applyEnforceBounds({
     proofRequest.request,
     selectedCredentials,
   );
-  let bounds = [];
 
-  descriptorBounds.forEach((items) => {
+  let skipProvingKey = false;
+
+  descriptorBounds.forEach((items, credentialIdx) => {
     items.forEach((bound, idx) => {
-      bounds.push(bound);
+      builder.enforceBounds(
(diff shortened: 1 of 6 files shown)


packages/wasm/src/services/credential/bound-check.ts
```diff
@@ -61,60 +62,38 @@ export function applyEnforceBounds({
...
-      skipProvingKey = true;
-    });
+  const descriptorBounds = pexToBounds(
+    proofRequest.request,
+    selectedCredentials,
+  );
+  let bounds = [];
+
+  descriptorBounds.forEach((items) => {
+    items.forEach((bound, idx) => {
```
cykoder, line 80: this is incorrect, the pextobounds result is split per credential. this wont work with multiple credentials.

descriptorBounds[0] corresponds to the first selected credential
descriptorBounds[1] corresponds to the second selected credential

cykoder (changes requested):
needs changing for multiple credentials

packages/wasm/src/services/credential/bound-check.ts
maycon-mello: yes, i've just pushed an update to support multiple credentials

maycon-mello: still finishing some tests

cykoder (dismissed):

cykoder (approved):

PR #270, opened 2024-06-11
performance improvements for ack

maycon-mello:
message ack is done after processing. The wallet doesn't need to wait for the ack to finish to import the credential

### DIFF
--- a/packages/core/src/message-provider.ts
+++ b/packages/core/src/message-provider.ts
@@ -144,11 +144,11 @@ export function createMessageProvider({
             did,
             messageIds,
           })
+          .then(() => {
+            logger.performance('Acknowledged messages', startTime);
+          })
           .catch(err => {
             console.error('Failed to ack messages', err.message);
-          })
-          .finally(() => {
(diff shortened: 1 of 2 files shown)


packages/core/src/message-provider.ts
```diff
@@ -135,20 +135,21 @@ export function createMessageProvider({
...
+        relayService
+          .ackMessages({
+            did,
+            messageIds,
+          })
+          .catch(err => {
+            console.error('Failed to ack messages', err.message);
+          })
+          .finally(() => {
+            logger.performance('Acknowledged messages', startTime);
```
mike-parkhill, line 151: this feels like it would be confusing because in the case of an error the log will show:

```
Failed to ack messages
Acknowledged messages
```

maycon-mello: good point, I think we should log that only if it succeed then

maycon-mello: i've pushed an update

mike-parkhill (approved):

PR #272, opened 2024-06-21
fix: use new credential witness

Jayteekay:


mike-parkhill (commented):
can we integration test this at all?

Jayteekay:
> can we integration test this at all?

We already have one but it's commented out because certs implementation is not merged yet, so it would fail.

mike-parkhill:
>  it's commented out because certs implementation is not merged yet, so it would fail.

can we add a comment to the Certs ticket to remind us to come back to un-comment that test?

maycon-mello (approved):

maycon-mello (approved):

PR #275, opened 2024-07-03
feat: add ability to exclude placeholder bounds in pexToBounds

Jayteekay:


mike-parkhill (commented):
feels like we should be able to unit tests these changes

maycon-mello (approved):

PR #279, opened 2024-07-29
DCKM-560: use ecosystem definition for biometric check credential

maycon-mello:


packages/react-native/lib/default-biometrics-plugin.ts
```diff
@@ -212,7 +212,7 @@ const issueBiometricMatchCredential = async enrollmentCredential => {
         // TODO: Certs template conditions needs to be updated to not require the data object
         // Workaround is to define that as true
```
mike-parkhill, line 213: comment should be updated

maycon-mello: done

mike-parkhill (approved):

mike-parkhill (approved):

PR #285, opened 2024-09-16
feat: Support injecting custom dataStore

maycon-mello:
After recent updates, TypeORM and SQL.js are causing a few issues when running the wallet-SDK in the browser. I'm moving TypeORM to a separate package so that we can inject lightweight dataStore impl for the browser browser. The cloud wallet code will be pushed as a separate PR, including its dataStore implementation for the browser. 

This PR also allows the client to inject a custom dataStore version into the Wallet SDK, which provides more flexibility.


integration-tests/helpers/wallet-helpers.ts
```diff
@@ -23,25 +26,40 @@ import {
...
   });
 
   didProvider = createDIDProvider({wallet});
   credentialProvider = createCredentialProvider({wallet});
   messageProvider = createMessageProvider({wallet, didProvider}) as any;
 
   console.log('Waiting for network connection');
-  await wallet.waitForEvent(WalletEvents.networkConnected);
+
+  if (dontWaitForNetwork) {
```
mike-parkhill, line 60: this seems backwards to me. The boolean is `dontWait` but then right away we wait.

packages/data-store-typeorm/src/migration/index.ts
```diff
@@ -18,7 +19,7 @@ const migrations = [
 
 export async function migrate({dataStore}: ContextProps) {
   // bootstrap v2 data
-  await bootstrapTables(dataStore.db);
+  await bootstrapTables(dataStore.db as any as DataSource);
```
mike-parkhill, line 22: why not use the getDataSource() helper here?

mike-parkhill (approved):

packages/data-store-typeorm/src/migration/index.ts
maycon-mello: good point, I've pushed an update

mike-parkhill (approved):

PR #291, opened 2024-10-11
Update docs-push.yml

mike-parkhill:


.github/workflows/docs-push.yml
```diff
@@ -1,19 +1,58 @@
+name: Push docs to knowledgebase
 on:
   workflow_dispatch:
-  
+  pull_request:
+  push:
+    branches: main
```
maycon-mello, line 6: We are using the `master` branch instead of `main`

mike-parkhill: fixed - think this is ready for re-review now

maycon-mello (approved):

PR #307, opened 2024-11-13
split credential and blockchain sdks

maycon-mello:


package.json
mike-parkhill, line 1: should @docknetwork/sdk be removed from here?

mike-parkhill (dismissed):

mike-parkhill (approved):

PR #310, opened 2024-11-28
cheqd migration

maycon-mello:


integration-tests/cheqd-credentials.test.ts
```diff
@@ -0,0 +1,59 @@
...
+import { CheqdCredentialNonZKP } from './data/credentials/cheqd-credentials';
+import { closeWallet, createNewWallet, getCredentialProvider, getWallet } from './helpers';
+import { ProofTemplateIds, createProofRequest } from './helpers/certs-helpers';
+
+describe('Cheq integration tests', () => {
+  beforeAll(async () => {
+    await createNewWallet();
+  });
+
+  it('should verify a non ZKP cheqd credential', async () => {
```
mike-parkhill, line 12: do we need one for ZKP too?

packages/cli/src/commands/ecosystem.ts
```diff
@@ -18,50 +18,50 @@ ecosystemCommands
...
     const didKeyPair = didKeyPairs[0];
 
-    await dockService.init({
+    await blockchainService.init({
       address: 'wss://knox-1.dock.io',
     });
 
-    await dockService.waitDockReady();
-    await dockService.ensureDockReady();
+    await blockchainService.waitDockReady();
```
mike-parkhill, line 25: might be an Oleg question, but do we need a `waitCheqdReady`?

packages/core/src/wallet-wasm.ts
```diff
@@ -1,49 +1,65 @@
...
+
   await keyringService.initialize({
     ss58Format: networkConfigs.addressPrefix,
   });
 
-  dockService
+  blockchainService
     .init({
       address: networkConfigs.substrateUrl,
+      cheqdApiUrl: networkConfigs.cheqdApiUrl,
```
mike-parkhill, line 60: are we mixing network configs all together? Would it make more sense to pass an array of config objects (one per network)?

packages/wasm/src/services/blockchain/service.ts
```diff
@@ -0,0 +1,209 @@
...
+} from '@docknetwork/dock-blockchain-modules';
+import {EventEmitter} from 'events';
+import {Logger} from '../../core/logger';
+import {once} from '../../modules/event-manager';
+import {utilCryptoService} from '../util-crypto';
+import {InitParams, validation} from './configs';
+
+let dockInstance;
+
+export function getDock() {
```
mike-parkhill, line 28: no getCheqd?

Would a look up by name be more extensible in the future? e.g. getChainService(chainname)

packages/core/src/wallet-wasm.ts
maycon-mello: for now yes, just because we need to support dock and cheqd at the same time. We still have multiple entries for, one for testnet and another for mainnet and the user can choose which one to use.

maycon-mello: if we had to support multiple chains connected at the same time in the long run, then I think it would make sense to refactor our network logic for doing that. But since its not a requirement for later, I decided to use the easier approach

packages/wasm/src/services/blockchain/service.ts
maycon-mello: I should remove this function, we will not be using dock directly anymore

packages/wasm/src/modules/network-manager.ts
```diff
@@ -14,11 +15,13 @@ export const SUBSTRATE_NETWORKS : Record<NetworkId, NetworkInfo> = {
     name: 'Dock PoS Mainnet',
     substrateUrl: ['wss://mainnet-node.dock.io', 'wss://mainnet-node-2.dock.io'],
     addressPrefix: 22,
+    cheqdApiUrl: 'https://cheqd-mainnet-rpc.autostake.com/',
```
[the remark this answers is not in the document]
maycon-mello: By default the wallet can't connect to multiple networks at the same time, the network resolver and switch logic would not work properly. To avoid a major refactoring I added substrate and cheqd into the same entry. And the wallet keep with the `mainnet`, and `testnet` networks as we had before.

integration-tests/cheqd-credentials.test.ts
maycon-mello: yes, would be good to add one here. I was planning to add zkp with accumulators, but makes sense to have zkp without accumulators too

packages/cli/src/commands/ecosystem.ts
maycon-mello: yes, its missing here. I will rename this method to be generic, and inside it will wait for both cheqd and substrate to be ready

integration-tests/cheqd-credentials.test.ts
maycon-mello: We are facing few issues with zkp verification, I will push that test as separate PR probably along with accumulators tests

mike-parkhill (approved):

PR #319, opened 2024-12-24
integration with Cheqd accumulators

maycon-mello:


mike-parkhill (commented):
can we add some integration tests with cheqd VC's?

packages/wasm/src/services/credential/index.test.js
```diff
@@ -401,7 +401,7 @@ describe('Credential Service', () => {
           },
         }),
       ).toBe(
-        '0xa632a41f2fbdb681c14b33daae4fcc46af41661b90b35c4ac1545c9bebf0d7cc',
+        'dock:accumulator:0xa632a41f2fbdb681c14b33daae4fcc46af41661b90b35c4ac1545c9bebf0d7cc',
```
mike-parkhill, line 404: do we need to support/test both formats (i.e. both with the prefix and without) so we ensure support of old credentials?

maycon-mello: the prefix didn't change in the credential, credentials always had prefix. The SDK used to require the registryId without prefix, so we had to remove that when generating the presentation. Now that's not needed anymore

packages/wasm/src/services/credential/service.ts
```diff
@@ -405,7 +405,7 @@ class CredentialService {
       if (witness) {
         const details = await getWitnessDetails(credential, witness);
         const chainModule =
-          credential.credentialStatus.id.indexOf('accumulator:dock:') === 0
+          credential.credentialStatus.id.indexOf('dock:accumulator') === 0
```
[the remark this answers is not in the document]
maycon-mello: `accumulator:dock:` doesn't exist. I've pushed that in an initial commit, but it was wrong.

mike-parkhill (approved):

PR #323, opened 2025-01-08
workaround for truvera cross-origin redirects

maycon-mello:


mike-parkhill (approved):

packages/react-native/lib/default-biometrics-plugin.ts
```diff
@@ -185,7 +185,8 @@ const issueBiometricMatchCredential = async enrollmentCredential => {
   let created = getIssuanceDate();
 
   if (schema?.id) {
-    const fetchSchemaResponse = await axios.get(schema?.id).catch((err) => {
+    // React native iOS can't handle cross-origin redirects
+    const fetchSchemaResponse = await axios.get(schema?.id?.replace('dock', 'truvera')).catch((err) => {
```
mike-parkhill, line 189: should we be more specific to avoid accidentally replacing it in the wrong spot? i.e. replace on schema.dock.io

maycon-mello: good point, let me update that logic

mike-parkhill (approved):

PR #330, opened 2025-01-22
upgrade sdk + remove old sdk references

maycon-mello:


packages/wasm/src/core/polkadot-utils.js
```diff
@@ -20,6 +18,19 @@ const polkadotTypesToKeys = {
   ecdsa: 'EcdsaSecp256k1VerificationKey2019',
 };
 
+export function getKeyPairType(key) {
+  const keyType = key.type || key.constructor.VerKeyType;
+  if (keyType) {
+    return keyType;
+  }
+
+  if (key instanceof Ed25519Keypair) {
```
mike-parkhill, line 27: we don't need to worry about BBS and BLS key types here?

setup-tests.js
```diff
@@ -38,23 +38,23 @@ require('@docknetwork/wallet-sdk-wasm/src/setup-tests');
...
-      addAttributeToReveal: mockAddAttributeToReveal,
-      createPresentation: mockCreatePresentation,
-      deriveCredentials: mockDeriveCredentials,
-      presBuilder: {
-        enforceBounds: jest.fn(),
-      },
-    };
-  });
-});
+// jest.mock('@docknetwork/sdk/presentation', () => {
```
mike-parkhill, line 41: do we still need this?

integration-tests/verification-flow/range-proofs.test.ts
```diff
@@ -49,92 +49,94 @@ describe('Range proofs verification', () => {
...
-      attributesToReveal,
-    });
-
-    const presentation = await controller.createPresentation();
-
-    expect(
-      presentation.verifiableCredential[0].credentialSubject,
-    ).toBeDefined();
-  });
+  // TODO: The DID for this credential is not available anymore
```
mike-parkhill, line 52: why isn't the DID available any more?

maycon-mello: I'm not sure yet; Agne shared This credential a few weeks ago. I will need to investigate a bit more

setup-tests.js
maycon-mello: I've just removed this commented code

packages/wasm/src/core/polkadot-utils.js
maycon-mello: this is only used to generate dids, so I think it should not be needed. It also matches with certs implementation

maycon-mello:
Had to disable the credit_score credential test, that's failing to generate presentations for that cred. Sales demo is currently failing, will try to get that fixed along with sales demo updates

mike-parkhill (approved):

PR #331, opened 2025-01-27
Fix pex to bounds to match the API

cykoder:


packages/wasm/src/services/credential/pex-helpers.js
```diff
@@ -49,23 +60,37 @@ function getAttributeName({field, selectedCredentials, index}) {
   return attributeName;
 }
 
-export function pexToBounds(
-  pexRequest,
-  selectedCredentials = [],
-  isRawBoundies = false,
```
cykoder, line 55: not sure what this was used for, but the callers to pexToBounds dont use it so assumed it can be safely removed

packages/wasm/src/services/credential/pex-helpers.js
```diff
@@ -49,23 +60,37 @@ function getAttributeName({field, selectedCredentials, index}) {
...
-) {
+export function pexToBounds(pexRequest, selectedCredentials = [], removeFromRequest = false) {
```
cykoder, line 63: API validation uses `removeFromRequest: true`, but i dont think that will effect the bounds. You shouldnt need to use it on the wallet side, for the API "mock wallet" we set to false

maycon-mello (approved):

PR #337, opened 2025-02-03
fix cheqd revocation

maycon-mello:


cykoder (approved):

packages/wasm/src/services/credential/bbs-revocation.ts
```diff
@@ -19,95 +19,33 @@ const trimHexID = id => {
...
-  return witness;
-}
-
 export const getWitnessDetails = async (credential, _membershipWitness) => {
   let witness = _membershipWitness;
+  let blockNo;
 
   try {
-    ({witness} = JSON.parse(_membershipWitness));
+    ({witness, blockNo} = JSON.parse(_membershipWitness));
```
gasher, line 27: I might be blind but I don't see this value being used.

packages/wasm/src/services/credential/bbs-revocation.ts
```diff
@@ -121,6 +59,18 @@ export const getWitnessDetails = async (credential, _membershipWitness) => {
 
   const membershipWitness = new VBMembershipWitness(hexToU8a(witness));
 
+  try {
+    await blockchainService.modules.accumulator.updateWitness(
+      registryId,
+      encodedRevId,
+      membershipWitness,
+      queriedAccumulator.created,
```
gasher, line 67: Shouldn't this use `blockNo` instead of `queriedAccumulator.created`? This might work for credentials that are issued with a brand new accumulator, but if it is using an existing accumulator that had some updates done to it the updates from `created` until `blockNo` shouldn't be applied to this.

PR #338, opened 2025-02-05
adding more logs to vc distribution flow

maycon-mello:


mike-parkhill (approved):
were you going to flip the log view? or is that in dock-app?

maycon-mello:
> were you going to flip the log view? or is that in dock-app?

that will be in the dock-app

PR #342, opened 2025-02-12
upgrade sdk + diable substrate connection on mainnet

maycon-mello:


packages/wasm/src/modules/network-manager.ts
```diff
@@ -13,7 +13,7 @@ export type NetworkId = 'mainnet' | 'testnet' | 'local' | 'custom';
 export const SUBSTRATE_NETWORKS : Record<NetworkId, NetworkInfo> = {
   mainnet: {
     name: 'Dock PoS Mainnet',
-    substrateUrl: ['wss://mainnet-node.dock.io', 'wss://mainnet-node-2.dock.io'],
+    substrateUrl: null,
```
mike-parkhill, line 16: the dependent code's smart enough to not choke when it sees a null here?

maycon-mello: Yes. We are using null on testnet already. And actually will need another ticket to perform a cleanup after the migration

maycon-mello: For now it allow us to rollback the change if needed, but later we can remove all substrate references.

mike-parkhill (approved):

cykoder (approved):

PR #345, opened 2025-02-21
Refactor presentation creation.

gasher:


packages/wasm/src/services/credential/service.ts
```diff
@@ -55,6 +56,150 @@ export function isAnnonymousCredential(credential) {
...
+      if (Array.isArray(filteredAttributes) && filteredAttributes.length > 0) {
+        await Promise.all(
+          filteredAttributes.map((attrib) =>
+            presentationInstance.addAttributeToReveal(idx, [attrib])
+          )
+        );
+      }
+      // Derive a W3C Verifiable Credential JSON from the above presentation
+      const credentials = await presentationInstance.deriveCredentials(presentationOptions);
+      presentation.addCredentials(credentials);
```
cykoder, line 177: i realised this is technically incorrect, it could change presentation order

packages/wasm/src/services/credential/service.ts
```diff
@@ -55,6 +56,150 @@ export function isAnnonymousCredential(credential) {
...
+  let pexRequiredAttributes = [];
+  if (pexForBounds?.request) {
+    pexRequiredAttributes = getPexRequiredAttributes(
+      pexForBounds.request,
+      vcsArray,
+    );
+  }
+
+  await Promise.all(
+    anoncredsVCs.map(async (vc) => {
```
cykoder, line 80: this will create all anoncreds derived credentials at once, which could work, but in a large presentation use case it might result in node timeout errors. might be better to do them sequentially

maycon-mello (approved):

maycon-mello:
The default presentation implementation already covers this work

PR #346, opened 2025-02-24
sqlite writing queue

maycon-mello:


packages/data-store-typeorm/src/entities/document/create-document.ts
```diff
@@ -19,64 +26,66 @@ export async function createDocument({
   json: any;
   options?: any;
 }): Promise<WalletDocument> {
-  logger.debug(`Creating document with id ${json.id}...`);
-  if (json.id) {
-    const existingDocument = await getDocumentById({
-      dataStore,
-      id: json.id,
-    });
+  return writeMutex.runExclusive(async () => {
```
mike-parkhill, line 29: we're sure this library gets translated properly when running on a device?

mike-parkhill: does the mutex get released automatically if there's an error?

maycon-mello: yes, I did few testing using iOS and android devices. didn't notice any issues on that

mike-parkhill (approved):

PR #350, opened 2025-03-04
DCKW-611 multiple keys can be registered to unlock the cloud wallet

gasher:


packages/core/src/cloud-wallet.ts
```diff
@@ -2,114 +2,337 @@ import {
...
+ * @param identifier User's identifier as additional entropy (email, phone number, etc.)
+ * @returns Keys for accessing the KeyMappingVault
+ */
+export async function deriveKeyMappingVaultKeys(
+  biometricData: any,
+  identifier: string
+): Promise<{ hmacKey: string; agreementKey: string; verificationKey: string }> {
+  const processedData = JSON.stringify(biometricData);
+
+  const saltBytes = new TextEncoder().encode(identifier);
```
cykoder, line 28: i would remove this custom key derivation which is just `SHA256({{SALT_STR}}{{BIOMETRIC_STR}})` with a package like https://www.npmjs.com/package/futoin-hkdf which implements the HKDF RFC properly

cykoder: with this we can specify the desired output length, which is useful because then instead of encrypting/decrypting the master key for storage you can just xor it like ```XOR(MASTER_KEY, BIOMETRIC_KEY)``` for storage. generally XOR'ing with same key size is acceptble for this case, its no more secure to "properly" encrypt/decrypt in this case because all code is open source/easily readable anyway - so the exact derivation/encryption can be worked out and applied to raw data or any derived keys. 

the comments in this thread should be applied to `deriveBiometricEncryptionKey` too. i do wonder if its worth having separate edv/key mapping keys derived from the same biometric - but if we do, it should flow more like:
```biometric input -> derive biometric key -> derive edv key from biometric key```

whereas right now its more like
```biometric input -> derive biometric key```
```biometric input -> derive edv key```

the benefit then is when you have an authentication mechanism which produces a secure key itself (rather than hashing a json string like with the biometric example) you can use that key directly to derive edv/encryption keys.

File: packages/core/src/cloud-wallet.ts [hunk 2]
```diff
@@ -2,114 +2,337 @@ import {
...
+  operations: QueuedOperation[];
+  isProcessing: boolean;
+}
+
+export async function generateCloudWalletMasterKey(): Promise<{ mnemonic: string; masterKey: string }> {
+  const mnemonic = await utilCryptoService.mnemonicGenerate(MNEMONIC_WORD_COUNT);
 
-  return base64url.encode(Buffer.from(masterKeyBytes));
+  const seedBytes = await utilCryptoService.mnemonicToMiniSecret(mnemonic);
+  const masterKey = base64url.encode(Buffer.from(seedBytes));
```
cykoder, line 320: its a waste of time/cpu to constantly encode/decode keys, just pass as buffer/u8 instead of strings

packages/core/src/cloud-wallet.ts
gasher: > which implements the HKDF RFC properly
Yeah I don't really know why I didn't use a library for this.
> generally XOR'ing with same key size is acceptble for this case, its no more secure to "properly" encrypt/decrypt in this case
That is true but there is a different reason why I went ahead with the encryption approach: AES-GCM includes an authentication tag that verifies the decryption was successful while XOR has no built-in way to verify if the result is correct. Using this we can have multiple biometrics registered to the same email (face and fingerprint) and we'll know which one the user is using just by trying to decrypt them and seeing which one works.

packages/core/src/cloud-wallet.ts
gasher: This function is user-facing and I don't think returning the master key as buffer is a good DX, but I agree there were some unnecessary encoding/decoding which I fixed

packages/core/src/cloud-wallet.ts
```diff
@@ -2,13 +2,302 @@ import {
...
+  identifier: string
+): Promise<string> {
+  const keyMappingEdv = await initializeKeyMappingVault(
+    edvUrl,
+    authKey,
+    biometricData,
+    identifier
+  );
+
+  const result = await keyMappingEdv.find({
```
cykoder, line 239: maybe not for this PR but we should have some local cache here, as this document wont change once created

cykoder (changes requested):

packages/core/src/cloud-wallet.ts
```diff
@@ -2,13 +2,302 @@ import {
...
+    equals: {
+      'content.id': identifier
```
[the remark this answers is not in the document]
gasher: The keys for the KeyMappingVault are derived from the biometric data and the email so there will be only one document in that vault (the encrypted key). There is no need to use a different identifier because it won't matter.

packages/core/src/cloud-wallet.ts
gasher: Will keep this in mind.

packages/core/src/cloud-wallet.ts
cykoder: dont have a major problem with using encryption/decryption instead of xor if it has some benefits, just thinking we could make the implementation "easier". i suppose its not very large data to encrypt/decrypt either so performance wont really be a concern on lower end mobile devices etc

packages/core/src/cloud-wallet.ts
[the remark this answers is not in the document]
cykoder: the EDV internally creates blind indexes based on the "content.id", im not 100% sure if that requires uniqueness even across multiple EDVs or not. still, it doesnt make much logical sense in my opinion to use an email as a content id. to conform with everything else we do, it should be a URI like suggested above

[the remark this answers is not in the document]
gasher: I doubt there are uniqueness requirements across EDVs since that will break incremental ids but I agree we should conform with everything else we do. Updated to use a DID

packages/wasm/src/services/edv/service.ts
```diff
@@ -130,6 +131,16 @@ export class EDVService {
     return { verificationKey, agreementKey, hmacKey };
   }
 
+  async generateContentId(key: Uint8Array) {
```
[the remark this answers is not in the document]
gasher: makes sense. thanks

cykoder (approved):

PR #353, opened 2025-03-20
drop did dock creation

maycon-mello:


packages/wasm/src/services/dids/index.test.js
mike-parkhill, line 1: some linting errors here

mike-parkhill (approved):

PR #354, opened 2025-03-31
feat: Support OID4VCI credential offer deep link

maycon-mello:


packages/wasm/src/services/credential/service.ts
```diff
@@ -199,7 +199,7 @@ class CredentialService {
   }) {
     const searchParams = new URL(uri).searchParams;
     const params = new URLSearchParams(searchParams);
-    const credentialOfferEncoded = params.get('credential_offer');
+    const credentialOfferEncoded = params.get('credential_offer') || params.get('credential_offer_uri');
```
mike-parkhill, line 202: can we test that this works somehow?

maycon-mello: yep, let me add proper tests here

cykoder: this logic is incorrect, `credential_offer_uri` is not an encoded credential offer. the lines immediately below it will fail. its a uri you have to request to get the offer JSON

packages/wasm/src/services/credential/index.test.js
```diff
@@ -442,4 +444,119 @@ describe('Credential Service', () => {
...
+      });
+
+      expect(result).toEqual({
+        credential: mockCredential,
+      });
+    });
+
+    it('should handle credential_offer_uri parameter', async () => {
+      const uri =
+        'openid-credential-offer://?credential_offer_uri=%7B%22credential_issuer%22%3A%22https%3A%2F%2Fapi-testnet.truvera.io%2Fopenid%2Fissuers%2F7eff516f-69fb-4b9d-94dc-e88308ec0c4c%22%2C%22credentials%22%3A%5B%22ldp_vc%3AMyCredential%22%5D%2C%22grants%22%3A%7B%22urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Apre-authorized_code%22%3A%7B%22pre-authorized_code%22%3A%22AAL4MPpEpWY6daBxVxJ8Q3chxOhvc9qIV3EAyj7dvps%22%2C%22user_pin_required%22%3Afalse%7D%7D%7D';
```
cykoder, line 509: this test is still incorrect, credential_offer_uri must be a valid resolveable URI. here its url encoded JSON

packages/wasm/src/services/credential/service.ts
cykoder: @maycon-mello why was this marked as resolved? i still see the same code/issues as the comment above

maycon-mello: I didn't see that comment I guess, let me handled that one too.

packages/wasm/src/services/credential/index.test.js
maycon-mello: can we assume the credential_offer_uri will return a json?

packages/wasm/src/services/credential/service.ts
```diff
@@ -199,23 +199,19 @@ class CredentialService {
...
       clientId: 'dock.wallet',
       authorizationRequest: {
         redirectUri: 'dock-wallet://credentials/callback',
         clientId: 'dock.wallet',
-        scope: credentialOffer.credentials[0],
       },
     });
 
+    const format = 'ldp_vc';
+    const credentialTypes = client.getCredentialsSupported()[0].scope.replace(`${format}:`, '');
```
cykoder, line 213: would be better to use `client.getCredentialsSupported()[0].scope.split(':')[1]` or similar, but also keep in mind that the scope parameter for 3d party implementations may have different formatting. so `ldp_vc:` is not guaranteed to exist. scope could just be like `AnotherCredentialType` or `sdjwt:AnotherCredentialType`

probably more robust to do:
```
const { scope }  = client.getCredentialsSupported()[0];
const scopeSplit = scope.split(':');
const credentialType = scopeSplit[scopeSplit.length - 1];
```

cykoder (approved):

PR #363, opened 2025-05-01
support multiple biometric providers

maycon-mello:


integration-tests/biometric-provider-truvera.test.ts
```diff
@@ -0,0 +1,296 @@
...
+      }
+      
+      // Remove any existing match credentials
+      const existingMatchCredentials = await credentialProvider.getCredentials(biometricConfigs.biometricMatchCredentialType);
+      for (const credential of existingMatchCredentials) {
+        await credentialProvider.removeCredential(credential.id);
+      }
+      
+      const proofRequest = {
+        input_descriptors: [
```
mike-parkhill, line 211: could this proof request be pulled into a constant? it's appears several times

integration-tests/biometric-provider-truvera.test.ts
```diff
@@ -0,0 +1,296 @@
...
+      // Wait for async operations to complete
+      await new Promise(resolve => setTimeout(resolve, 100));
+      
+      // Simulate the completion event ourselves (in real app, the wallet would do this)
+      const walletDID = await getDIDProvider().getDefaultDID();
+      const result = await idvProvider.match(walletDID, enrollmentCredentials[0], proofRequest);
+      biometricProvider.eventEmitter.emit('onComplete', result);
+      
+      // Wait for startIDV to finish
+      await startIDVPromise;
```
mike-parkhill, line 239: feels like the lines above this are repeated - any chance to pull out a helper method?

packages/react-native/lib/trustx-biometric-plugin.test.ts
```diff
@@ -0,0 +1,336 @@
...
+    "biometric_enrollment_id": "enrollment-123",
+    "created": "2023-01-01T00:00:00Z"
+  },
+  "proof": {
+    "type": "Ed25519Signature2018",
+    "proofValue": "mock-proof-value"
+  }
+};
+
+const mockMatchCredential = {
```
mike-parkhill, line 54: this is the credential being issued by Daon? Do we need to worry about them changing the schema ever?

packages/react-native/lib/trustx-biometric-plugin.ts
```diff
@@ -0,0 +1,79 @@
...
+  configs,
+}: {
+  wallet: IWallet;
+  eventEmitter: EventEmitter;
+  configs: TrustXIDVConfig;
+}): IDVProvider {
+  return {
+    async enroll(walletDID, proofRequest) {
+      
+      // TODO: get start URL from TrustX API
```
mike-parkhill, line 32: when can we come back for this TODO?

mike-parkhill (commented):
should we be looking at adding a status check for the Trustx environment to ensure it's available and performing as expected?

packages/react-native/lib/trustx-biometric-plugin.ts
maycon-mello: need to remove this comment, it was already done

mike-parkhill (approved):

packages/core/src/credential-provider.ts
```diff
@@ -275,6 +276,42 @@ async function syncCredentialStatus({
...
+  
+  // Remove the main credential document
+  await wallet.removeDocument(credentialId);
+  
+  // Remove related documents
+  try {
+    // Remove the witness document if it exists
+    await wallet.removeDocument(`${credentialId}#witness`);
+  } catch (err) {
+    // Witness document might not exist, which is fine
```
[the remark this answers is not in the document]
mike-parkhill: Of course, the flip side is if the look up is more expensive than the exception generation, but I don't think that's the case here since it's all local

PR #369, opened 2025-05-21
fixing issue with biometric credential issuance

maycon-mello:


packages/react-native/lib/truvera-biometric-plugin.ts
```diff
@@ -30,6 +30,12 @@ export const getIssuanceDate = () => {
   return dateObject.toISOString();
 };
 
+function getBiometricCreatedDate() {
```
mike-parkhill, line 33: this could probably be more generic if you pass the datetime in and convert it

```
function convertDateTimeToDate(dt) {
  return dt.split('T')[0];
}
```

mike-parkhill (approved):

mike-parkhill (approved):

PR #371, opened 2025-05-27
ignore examples folder in docs push action

maycon-mello:


mike-parkhill (approved):
We should probably look at the underlying dependabot alerts though too, right?

PR #389, opened 2025-06-12
Update WalletSDK examples to use the latest React Native version 0.79

maycon-mello:


examples/react-native-example/README.md
mike-parkhill, line 1: this definitely feels like it could use a little more added to it, but if these commands are enough to get it going that's probably a good enough start.

mike-parkhill (approved):

mike-parkhill (approved):

mike-parkhill (approved):

PR #404, opened 2025-07-17
improve pexToBounds documentation + adding more tests for range proofs with KVAC variation

maycon-mello:
…s with KVAC variation

### DIFF
--- a/integration-tests/verification-flow/range-proofs.test.ts
+++ b/integration-tests/verification-flow/range-proofs.test.ts
@@ -7,9 +7,15 @@ import {
 import {createVerificationController} from '@docknetwork/wallet-sdk-core/src/verification-controller';
 import {createProofRequest} from '../helpers/certs-helpers';
 import axios from 'axios';
-import {boundCheckSnarkKey} from './proof-requests';
 
 describe('Range proofs verification', () => {
+  let proofRequest: any;
+
+  beforeAll(async () => {
+    proofRequest = await createProofRequest(
+      'db3ca660-f353-4298-8522-73ef95ea1a16',
+    );
+  });
   it('should create range proofs presentation without subject', async () => {
     const wallet: IWallet = await getWallet();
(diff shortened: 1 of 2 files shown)


gasher (dismissed):

integration-tests/verification-flow/range-proofs.test.ts
```diff
@@ -49,94 +51,45 @@ describe('Range proofs verification', () => {
...
+    );
+
+    getCredentialProvider().addCredential(credential);
 
+    await controller.start({
+      template: proofRequest.qr,
+    });
+
+    // Reveal only dateOfBirth attribute
+    const attributesToReveal = [
```
mike-parkhill, line 74: is having the word, "Reveal" here confusing if we're trying to prove that the attribute isn't actually revealed?

maycon-mello: Good point, as range proof attribute will not be revealed in the presentation we could skip this part. pexToBounds will define which range proof attributes should be included, and any range proof attributes added to this attributesToReveal array will be removed automatically

maycon-mello: I've updated the tests and also added better assertions for range proof bounds

integration-tests/verification-flow/range-proofs.test.ts
```diff
@@ -49,6 +52,19 @@ describe('Range proofs verification', () => {
     expect(
       presentation.verifiableCredential[0].credentialSubject,
     ).toBeUndefined();
+
+    // assert that the bounds for dateOfBirth are present
```
mike-parkhill, line 56: should we also assert that the dateOfBirth isn't explicitly there?

maycon-mello: I think we already do that when checking the credentialSubject is empty

```

mike-parkhill (approved):

PR #410, opened 2025-08-22
adding support for did service endpoints

maycon-mello:


packages/relay-service/src/index.js
```diff
@@ -39,6 +45,33 @@ const sendMessage = async ({keyPairDoc, recipientDid, message, type}) => {
     keyAgreementKey,
   });
 
+  if (useDIDServiceEndpoint) {
```
cykoder, line 48: i would suggest that if a did document has a service endpoint, it should always be used for didcomm instead of a relay (below this check)

maycon-mello: sure that makes sense. We have the wallet to wallet verification that uses didcomm messages but it is using did:key only, so i guess it should be safe

maycon-mello: I've pushed a commit for this update

packages/relay-service/src/index.js
```diff
@@ -39,6 +39,30 @@ const sendMessage = async ({keyPairDoc, recipientDid, message, type}) => {
...
+  if (serviceEndpoint) {
+    try {
+      const result = await axios.post(serviceEndpoint.uri, jweMessage, {
+        headers: {
+          'Content-Type': 'application/didcomm-encrypted+json',
+          Accept: 'application/didcomm-encrypted+json',
+        },
+      });
+
+      return result.data;
```
cykoder, line 59: returning result.data here and again below will produce two different results for success. should ideally return the same object or  just a success value if its not actually needed

packages/relay-service/src/index.js
```diff
@@ -39,6 +39,30 @@ const sendMessage = async ({keyPairDoc, recipientDid, message, type}) => {
     keyAgreementKey,
   });
 
+  const didDocument = await blockchainService.resolveDID(recipientDid);
+  const service = didDocument.service.find(
+    endpoint => endpoint.type === 'DIDCommMessaging',
+  );
+
+  const serviceEndpoint = service?.serviceEndpoint[0];
```
cykoder: you should loop all service endpoints until one of them is successful and has the "accept" type that you are sending with, i.e `didcomm/v2`.

Rank candidates (typical policy)
- Prefer entries advertising DIDComm v2 and HTTPS
- Respect the endpoint’s own accept order (sender should pick the most preferred media type it can satisfy).
- If there are multiple DIDCommMessaging services, prefer them in the order they appear / are advertised (common practice for “preference”).

cykoder: also you cant assume serviceEndpoint is an array, see https://identity.foundation/didcomm-messaging/spec/#service-endpoint

it could be an object too (typical for jsonld)

mike-parkhill (approved):

PR #416, opened 2025-09-15
Dcka 4877 setup up js doc builds in ci cd

maycon-mello:


packages/core/src/types.ts
```diff
@@ -5,12 +5,50 @@ import {
 } from '@docknetwork/wallet-sdk-data-store/src/types';
 import {EventEmitter} from 'events';
 
+/**
+ * Legacy V1 wallet interface for backward compatibility
```
mike-parkhill, line 9: are we okay with devs referencing this, or should we mark it as obsolete or something so they don't start adding dependencies on it?

packages/core/src/types.ts
```diff
@@ -19,55 +57,213 @@ export interface IV1Wallet {
...
+   * @param {any} json - The encrypted wallet JSON
+   * @param {string} password - Password to decrypt the wallet
+   * @returns {Promise<void>}
+   * @see {@link https://w3c-ccg.github.io/universal-wallet-interop-spec/}
    */
   importUniversalWalletJSON: (json: any, password: string) => Promise<void>;
+  
+  /**
+   * Decrypts and retrieves documents from an encrypted wallet without importing
+   * @param {any} json - The encrypted wallet JSON
```
mike-parkhill, line 208: is this also expecting the universal wallet format? Should we add a `@see` reference here too?

mike-parkhill (approved):

PR #429, opened 2025-10-03
fixing knowledge base docs integration

maycon-mello:


docs/getting-started/README.md
```diff
@@ -0,0 +1,121 @@
+# Getting Started
```
esplinr, line 1: Was it an accident that you added this file back? My PR moved it from this subdirectory to the docs folder named getting-started.md.

.github/workflows/docs-push.yml
```diff
@@ -34,8 +34,9 @@ jobs:
       - name: Copy files
         run: |
           # Specify the files or directories you want to copy
-          cp -r docs/* knowledge-base/developer-documentation/wallet-sdk/
-          cp README.md knowledge-base/developer-documentation/wallet-sdk/README.md
+          rm -rf ./knowledge-base/credential-wallet/wallet-sdk
+          rsync -av docs/ ./knowledge-base/credential-wallet/wallet-sdk/
```
esplinr, line 38: If you need to exclude the API docs from the deletion, then I recommend using rsync --delete --exclude . Then you don't need a separate rm command.
If you don't need to exclude the API docs from the deletion, then you can use rm and cp and you don't need rsync.

maycon-mello:
Thanks @esplinr. I've handled your PR feedback. Please have another look.

PR #502, opened 2026-04-16
dcka 5447 wallet sync issues

maycon-mello:
- improve error handlers for cloud wallet

### DIFF
--- a/packages/wasm/src/services/edv/service.ts
+++ b/packages/wasm/src/services/edv/service.ts
@@ -251,20 +251,47 @@ export class EDVService {
   }
 
   /**
-   * Finds documents in the EDV based on query parameters
-   * @param {Object} params - Query parameters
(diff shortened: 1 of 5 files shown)


packages/wasm/src/services/edv/service.ts
```diff
@@ -251,20 +251,47 @@ export class EDVService {
...
    * });
    */
-  find(params: any) {
-    return this.storageInterface.find(params);
+  async find(params: any) {
+    try {
+      return await this.storageInterface.find(params);
+    } catch (error) {
+      if (
+        error.message.includes('Vault indices') &&
```
mike-parkhill: why is the check split in two?

maycon-mello: Actually, it was supposed to check for `Vault indices` and `not exist`.

mike-parkhill (approved):

PR #507, opened 2026-04-30
dcka 3147 wallet delegation policy

maycon-mello:
- wallet delegation policy


integration-tests/delegation/delegation-fixtures.ts
```diff
@@ -0,0 +1,367 @@
...
+              const: true,
+            },
+            capability: 'Reserve Flights',
+          },
+        ],
+      },
+      {
+        label: 'Booking Executor',
+        roleId: 'd39f29c4-fc3e-4b5d-9eae-9f576576e4fb',
+        attributes: ['*'],
```
mike-parkhill, line 291: do we have any examples where attributes are restricted?

packages/core/src/delegation/delegation-chain.ts
```diff
@@ -0,0 +1,50 @@
...
+
+type DelegationDetails = {
+  yourRole: string;
+  expires: Date;
+  maxDelegationDepth: number;
+  currentDelegationDepth: number;
+  capabilities;
+};
+
+export async function addDelegationChain(credential, delegationChain, wallet) {
```
mike-parkhill, line 16: does adding a new credential to the chain require removing and re-saving the whole chain again?

maycon-mello: No, this chain would not be changed. In case we re-delegate, a new chain will be created and sent along with the new credential.

integration-tests/delegation/delegation-fixtures.ts
maycon-mello: Yeah would be good to have some variants. I was planning to add those tests to the redelegation work. This current PR is primarily to add support for the root credential and to fetch the policy for rendering. In the re-delegation tests, we will need to ensure that restricted attributes are not visible.

mike-parkhill (approved):

PR #529, opened 2026-05-15
fix/sdjwt import over oid4vc

maycon-mello:
- feat: support importing decoded SD-JWT VC payloads
- honor issuer-advertised format in OID4VCI credential acquisition

### DIFF
--- a/packages/wasm/src/services/credential/oid4vci.js
+++ b/packages/wasm/src/services/credential/oid4vci.js
@@ -6,7 +6,10 @@ export function resolveOfferedCredentialConfig(client) {
     [];
 
   if (Array.isArray(supported)) {
-    return supported[0];
+    const matchedLegacy = supported.find(
+      entry => entry?.id && offerIds.includes(entry.id),
+    );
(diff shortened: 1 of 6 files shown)


packages/wasm/src/services/credential/oid4vci.js
```diff
@@ -0,0 +1,41 @@
+export function resolveOfferedCredentialConfig(client) {
+  const supported = client.getCredentialsSupported();
+  const offerIds =
+    client.credentialOffer?.credential_offer?.credential_configuration_ids ??
+    client.credentialOffer?.credential_configuration_ids ??
+    [];
+
+  if (Array.isArray(supported)) {
+    return supported[0];
```
mike-parkhill: how do we know the first entry in the array is the right one?

maycon-mello: good point, right now our API can only issue one format and this array is only used in the legacy flow. The latest OpenID specification will auto-select the proper option. But it's worth adding a filter here just to be safe

mike-parkhill (approved):

PR #537, opened 2026-05-29
delegation offer over didcomm 

maycon-mello:


packages/core/src/delegation/delegation-offer.ts
```diff
@@ -0,0 +1,450 @@
...
+  delegationPolicy: any;
+  delegationRole: string;
+  credentialId?: string;
+  expiresInMs?: number;
+}) {
+  if (credentialId) {
+    const parentCredential = await wallet.getDocumentById(credentialId);
+    if (parentCredential) {
+      assert(
+        isDelegatableCredential(parentCredential),
```
mike-parkhill, line 86: does this assert that the constraints/policies are being enforced? Or, where is that happening?

packages/core/src/delegation/delegation-offer.ts
```diff
@@ -0,0 +1,450 @@
...
+  const delegationOfferMessage = {
+    type: OOB_INVITATION,
+    id: delegationOffer.id,
+    from: issuerDID,
+    body: {
+      goal_code: GOAL_CODE,
+      goal:
+        goal ??
+        (finalIssuerName
+          ? `${finalIssuerName} is offering you a delegation`
```
mike-parkhill: should this text come out of translations?

packages/core/src/delegation/delegation-offer.ts
maycon-mello: Only the app is enforcing that, but I think it makes sense to enforce it here too, since we'll support delegation directly from the SDK. Let me add that.

mike-parkhill: Yeah, I think the SDK needs to enforce that - we won't be able to trust 3rd party tools to do it.

mike-parkhill (approved):

PR #555, opened 2026-07-10
Fixing npm build issues with 1.9.0 wasm package

maycon-mello:


packages/wasm/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@docknetwork/wallet-sdk-wasm",
-    "version": "1.9.0",
+    "version": "1.9.1-alpha.0",
```
mike-parkhill: why are we using an alpha tag here?

maycon-mello: This one was for a test build. The latest commit should have the proper package bump across the other packages

mike-parkhill (approved):

PR #557, opened 2026-07-17
feat/delegatable revocation dcka 5451

maycon-mello:


integration-tests/delegatable-revocation.test.ts
```diff
@@ -0,0 +1,130 @@
...
+
+function evictStatusListCache(credential: any) {
+  const url = credential?.credentialStatus?.statusListCredential;
+  if (url) cachedUris.delete(url.endsWith('/') ? url.slice(0, -1) : url);
+}
+
+const SPONSOR_KEY = process.env.TRUVERA_API_SPONSOR_KEY;
+const API_URL = process.env.DELEGATION_REVOCATION_API_URL;
+
+const ROLE_ID = 'e79c0d16-8739-4e54-94d7-53d9f1c97c71';
```
mike-parkhill, line 27: is this ever liable to change? I'm assuming it relates to a particular row in the staging db - which account does it belong to?

maycon-mello: This is from our mocks, not related to staging db

packages/core/src/delegation/delegation-issuance.ts
```diff
@@ -14,13 +19,17 @@ import {getAllDIDs, getDIDKeyPair} from '../did-provider';
  * @param delegationPolicy
  * @param delegationRoleId
  * @param rootCredentialId
+ * @param revocationContext when provided, a StatusList2021Entry is allocated and
+ *   embedded automatically so the credential can be revoked later. The wallet's
+ *   revocation registry index counter is advanced and persisted as a side effect.
```
mike-parkhill: terminology quibble, can we say the "counter is incremented" rather than "advanced"?

integration-tests/delegatable-revocation.test.ts
maycon-mello: The delegation policy gets embbeded into the credential, so it does not depends on workspace

packages/core/src/delegation/delegation-revocation.ts
```diff
@@ -0,0 +1,302 @@
...
+/**
+ * Common context for every signed revocation API call.
+ *
+ * `apiUrl` doubles as the JWT `aud` claim and MUST exactly match the server's
+ * configured SERVER_URL (no trailing slash) or the token is rejected with 401.
+ */
+export interface RevocationContext {
+  wallet: IWallet;
+  issuerDID: string; // the did:key that owns the registry / signs the JWT
+  truveraApiSponsorKey: string;
```
mike-parkhill, line 58: If we ever have to rotate the sponsor key will the registry still be usable?

maycon-mello: Yes, the sponsor key is not related to revocation registry. Its a top level auth middleware. The revocation registry authentication is handled by the did document owned by the wallet holder

mike-parkhill (commented):
I'm not seeing you working with accumulators in this - do we support anonymous delegation and revocation?

maycon-mello:
> I'm not seeing you working with accumulators in this - do we support anonymous delegation and revocation?

This is an off chain approach, so accumulators are not used. It will work for bbs properly, but KVAC is not supported yet
https://dock-team.atlassian.net/browse/DCKA-5481

mike-parkhill (approved):
Chatted in Slack about the anonymous revocation question. This will work with BBS, but not KVAC. See DCKA-5481 for that work.

PR #564, opened 2026-07-26
Fix SD-JWT signature verification

cykoder:


packages/wasm/package.json
```diff
@@ -23,14 +23,15 @@
...
+        "@docknetwork/credential-sdk": "0.56.2",
+        "@docknetwork/crypto-utils": "0.2.2",
         "@docknetwork/universal-wallet": "^2.0.1",
         "@docknetwork/vc-delegation-engine": "1.0.4",
         "@docknetwork/wallet-sdk-dids": "^1.9.1",
         "@noble/hashes": "1.8.0",
         "@scure/bip39": "^1.6.0",
-        "@sd-jwt/crypto-nodejs": "^0.15.0",
-        "@sd-jwt/sd-jwt-vc": "^0.15.1",
+        "@sd-jwt/crypto-nodejs": "0.9.2",
```
maycon-mello: is there any issue with using 0.15.0?

cykoder: it was just for testing for compat across platforms, will revert

packages/wasm/src/services/credential/sd-jwt.ts
```diff
@@ -93,7 +106,7 @@ export async function verifySDJWT(jwt, requiredAttribs?, options = {}) {
     };
   }
 }
-
+fix
```
mike-parkhill: is this meant to be in a comment?

cykoder: nope, shouldnt be there at all, guess i started typing there instead of git

maycon-mello (approved):

PR #566, opened 2026-07-28
persist the delegated credential in the offer for revocation purposes

maycon-mello:


packages/core/src/delegation/delegation-offer.ts
```diff
@@ -380,6 +378,10 @@ export const DELEGATION_REQUEST_HANDLER = {
       },
     });
 
+    delegationOffer.credential = delegatedCredential;
```
maycon-mello: we could persist the credentialStatus and credentialId, that should be enough for revocation. But I wonder if we might want to re-send a credential? so the full credentlai might be useful

packages/core/src/delegation/delegation-offer.ts
```diff
@@ -380,6 +380,12 @@ export const DELEGATION_REQUEST_HANDLER = {
       },
     });
 
+    // Persist revocation data on the delegation offer so the issuer can update it when revoking.
+    delegationOffer.credentialStatus = delegatedCredential.credentialStatus;
+    delegationOffer.credentialId = delegatedCredential.id;
+
+    await wallet.updateDocument(delegationOffer);
```
mike-parkhill, line 389: how heavy is the delegation offer? Do we need the full offer, or should we store a lighterweight document for this metadata?

maycon-mello: we keep the delegation policy used for the issuance, that's all. It would be needed for rendering purposes so you know what that credential is allowed to do.

packages/core/src/delegation/delegation-offer.ts
```diff
@@ -381,8 +381,10 @@ export const DELEGATION_REQUEST_HANDLER = {
     });
 
     // Persist revocation data on the delegation offer so the issuer can update it when revoking.
-    delegationOffer.credentialStatus = delegatedCredential.credentialStatus;
-    delegationOffer.credentialId = delegatedCredential.id;
+    delegationOffer.revocationData = {
```
mike-parkhill, line 384: Did you add tests to validate this is being stored? If so, I'd expect them to fail after this refactor

maycon-mello: We have revocation tests which is calling the revocation helpers directly and has access to the full credential. I just updated the test to ensure we assert the revocation data is there.

mike-parkhill (approved):

PR #572, opened 2026-07-31
dcka 5521 enforce https oid4vci

maycon-mello:
- enforce https for OID4VCi

### DIFF
--- a/packages/wasm/src/services/credential/index.test.js
+++ b/packages/wasm/src/services/credential/index.test.js
@@ -491,6 +491,11 @@ describe('Credential Service', () => {
         getCredentialsSupported: jest
           .fn()
           .mockReturnValue([{scope: 'ldp_vc:MyCredential'}]),
+        getIssuer: jest
+          .fn()
+          .mockReturnValue(
+            'https://api-testnet.truvera.io/openid/issuers/7eff516f-69fb-4b9d-94dc-e88308ec0c4c',
+          ),
       };
 
       jest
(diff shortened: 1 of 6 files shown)


packages/wasm/src/services/credential/index.test.js
```diff
@@ -570,6 +585,91 @@ describe('Credential Service', () => {
...
+          .spyOn(OpenID4VCIClientV1_0_13, 'fromURI')
+          .mockResolvedValue(mockClient);
+        jest
+          .spyOn(didService, 'createSignedJWT')
+          .mockResolvedValue('mock.jwt.token');
+
+        const result = await service.acquireOIDCredential({
+          uri,
+          holderKeyDocument: mockHolderKeyDocument,
+          allowInsecureHttpRequests: true,
```
mike-parkhill, line 667: when would we want to support insecure endpoints? During local dev?

maycon-mello: Yes, exactly — local dev / testing against an issuer running over plain HTTP (e.g. localhost or an internal staging box without TLS). It's opt-in via allowInsecureHttpRequests and defaults to false, so production stays HTTPS-only.

maycon-mello: yes, for local development it might be required. In the first try it will throw in the mobile app, the user might have the option to proceed with insecure requests, a new call will be made using this allowInsecureHttpRequests flag

mike-parkhill: we'd prompt them in the app to allow insecure connections or something? Similar to how browsers do it?

maycon-mello: Correct

mike-parkhill (approved):

PR #573, opened 2026-08-04
feature/allow custom local storage path on nodejs

maycon-mello:
- custom path for local storage in nodejs

### DIFF
--- a/examples/nodejs-example/readme.md
+++ b/examples/nodejs-example/readme.md
@@ -59,6 +59,16 @@ npm run oid4vc-example <OID4VC_URL>
 
 The example above creates an instance of the Wallet SDK, imports the credential into the wallet, and then logs the list of credentials available in the user database. Note that the imported credential is stored locally on the device, and in this Node.js example, it uses SQLite for storage.
 
+## Storage location in Node.js
+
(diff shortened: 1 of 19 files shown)


packages/wasm/src/setup-nodejs.js
```diff
@@ -2,7 +2,7 @@ import {LocalStorage} from 'node-localstorage';
 import {getRpcClient, initRpcClient} from './rpc-client';
 import rpcServer from './rpc-server';
 
-global.localStorage = new LocalStorage('./local-storage');
+global.localStorage = new LocalStorage(process.env.LOCAL_STORAGE_PATH || './local-storage');
```
mike-parkhill: Should we check for a `globalThis.localStorage` option before creating a new one here? Just comparing this path to the one in service.ts and it doesn't feel consistent.

examples/nodejs-example/readme.md
```diff
@@ -59,6 +59,16 @@ npm run oid4vc-example <OID4VC_URL>
 
 The example above creates an instance of the Wallet SDK, imports the credential into the wallet, and then logs the list of credentials available in the user database. Note that the imported credential is stored locally on the device, and in this Node.js example, it uses SQLite for storage.
 
+## Storage location in Node.js
```
mike-parkhill, line 62: should these notes be added to a Readme under the wasm package too?

mike-parkhill (commented):
A couple of linting issues

packages/wasm/src/setup-nodejs.js
maycon-mello: This file is used for nodejs unit tests only, will never be used in browsers. And actually knowing that the service.ts can now handle localStorage for nodejs we might want to remove this line. Let me try that out

mike-parkhill (approved):

PR #574, opened 2026-08-04
Dcka 5463 delegation request

maycon-mello:


packages/core/src/delegation/delegation-request.ts
```diff
@@ -0,0 +1,456 @@
...
+  wallet: any;
+}): Promise<void> {
+  const storedRequest = await wallet.getDocumentById(delegationRequest.id);
+  assert(
+    !!storedRequest,
+    `DelegationRequest ${delegationRequest.id} not found`,
+  );
+  await wallet.updateDocument({
+    ...storedRequest,
+    status: 'rejected',
```
mike-parkhill, line 345: do we need some sort of `reason` field to be tracked as to why it was rejected?

mike-parkhill (approved):

PR #578, opened 2026-09-15
fixing esm errors related to universal wallet

maycon-mello:
- fixing more esm errors related to universal wallet


mike-parkhill (approved):
I like it when we delete shims and hacks. 😄

